PR #15, opened 2019-02-27
introduce SearchBar

tempelgogo:
## Description
introduce SearchBar 

## Motivation & Context
vue ds workflow

refs #92 

## screenshot

Screenshot from 2019-03-25 21-13-20


DeepDiver1975:
Please have a look at the uikit classes https://getuikit.com/docs/navbar and see if the owncloud theme already holds valuable adjustments. THX

DeepDiver1975:
I removed the top bar because we have the top bar already in #87 
but the search component starting point can be reused

PR #62, opened 2019-03-18
Introduce oc-breadcrumbs

DeepDiver1975:


felixheidecke:
About the warning. We would use Vue-Router to … well … route. The router is not part of ODS which is no understanding. @DeepDiver1975 Do you have a solution? Should we just add it?

PR #85, opened 2019-03-21
icons: add file type icons from owncloud/core

DeepDiver1975:
Screenshot_2019-03-21 ownCloud Design System(5)


michaelstingl:
@DeepDiver1975 Why 5 different colors? I thought we can do this with theming when we use SVG (CSS).

DeepDiver1975:
Color of Icons is driven by a property of ic-icon components.
I'm showing them Here in different colors to be sure coloring is working in them (With exception of PDF and office Docs)

PR #87, opened 2019-03-22
oc-topbar: initial proposal

DeepDiver1975:
Screenshot_2019-03-22 ownCloud Design System


DeepDiver1975:
get component layout ideas from https://vuikit.js.org/guide/navbar

PR #104, opened 2019-03-26
Feature/automplete improvements

felixheidecke:
Add loader and overflow indicator to autocomplete


felixheidecke:
I see we need a different approach here, the suggestions need to load instead of blocking input. I'll handle that.

PR #105, opened 2019-03-26
Bugfix/fix search bar

DeepDiver1975:
Screenshot_2019-03-26 ownCloud Design System

### DIFF
--- a/src/patterns/OcSearchBar.vue
+++ b/src/patterns/OcSearchBar.vue
@@ -17,7 +17,7 @@
         <div
           uk-close
           v-if="query.length > 0"
-          @click="query = ''"
+          @click="clearInput"
           class="oc-search-clear uk-position-small uk-position-center-right"
         />
       </div>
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
> UX: This needs a send/search button

agreed - I failed to include a button :see_no_evil: - help appreciated ... feel free to commit into this branch - THX

felixheidecke:
I'd also move the search icon to the left and replace it with the loading spinner if it's loading. Then we have enough clutter free space for an actual button on the right.

felixheidecke:
Having the search query to the right increases the risk of running into width problems where we would need to ellipse the query on adjacent objects or smaller horizontal width.

DeepDiver1975:
@felixheidecke 
- [ ] clearing input does not emit the change event
- [ ] we might want an option to remove the button
Screenshot from 2019-03-27 12-56-35

felixheidecke:
There is the `button="false"` option to get rid of the button.

Issue #103, opened 2019-03-25
element: linear spinner


DeepDiver1975:
## refs
see https://vuetifyjs.com/en/components/progress#indeterminate

## behavior
- a spinner which is design like the progress bar

PR #107, opened 2019-03-26
Add OcLoader

felixheidecke:
Closes: https://github.com/owncloud/owncloud-design-system/issues/103

### DIFF
--- a/src/elements/OcLoader.vue
+++ b/src/elements/OcLoader.vue
@@ -1,0 +1,29 @@
+<template>
+  <div class="oc-loader" :aria-label="ariaLabel"></div>
+</template>
+<script>
+/**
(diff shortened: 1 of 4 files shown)


felixheidecke:
I also added the _Loader to Loading_ demo in the pattern library. @DeepDiver1975 Not sure if we need this?!

DeepDiver1975:
> I also added the _Loader to Loading_ demo in the pattern library. @DeepDiver1975 Not sure if we need this?!

this can be of great help in some cases - THX

DeepDiver1975:
Question is where to put such stuff ... this is more example code ... hmmm

I see these options:
- add this as template
- add this to the components oc-loader and oc-progress

DeepDiver1975:
>     * add this as template

@felixheidecke let's move this to a new template for now - THX

PR #109, opened 2019-03-26
add file actions and items

butonic:
This is a first approach:

Auswahl_013

Several questions:
1. I combined an `offcanvas`, `nav` and a `span` for this. It does slide in from the left, but we could use `mode: none` to make it just appear. The css for sliding the div in from the bottom eludes me. I already added css to make the canvas appear at the bottom, see [the upstream issue comment](https://github.com/uikit/uikit/issues/3445#issuecomment-437058627). Not the cleanest approach. And it is not modal ... if we even want that.
2. Does using an offcanvas make it easier for mobile navigation?
3. What about the label?
  a. How do we combine the text with the filename? I currently use a simple string replace.
  b. How do wo translate the "Open %s in" string?
4. How do we always close the action items offcanvas when an action item is clicked. Currently, I use ` uk-toggle="target: #file-actions"` in the item to reference the offcanvas, but that hardcodes an id into the element. We could also move that toggle to the template, but I guess there is a proper way to do that (via js, but a naive `UIkit.offcanvas("#file-actions").hide()` says it can't find UIkit and it would hardcode the id as well)?

### DIFF
--- a/src/patterns/OcFileActions.vue
+++ b/src/patterns/OcFileActions.vue
@@ -1,0 +1,179 @@
+<template>
+  <div :id="id" uk-offcanvas="mode: slide" class="uk-offcanvas uk-offcanvas-bottom">
+    <div class="uk-offcanvas-bar">
+      <span v-text="_label"></span>
+      <ul class="uk-nav">
+        <template v-for="(action, i) in actions">
+          <li :key="i">
+            <a class="uk-inline" @click="selectAction(action)">
+              <oc-icon v-if="action.icon" :name="action.icon" />
+              <span class="uk-text-top" v-text="action.label" />
+            </a>
+          </li>
+        </template>
+      </ul>
+    </div>
+  </div>
+</template>
+
+<style>
+.uk-offcanvas.uk-offcanvas-bottom {
+  left: auto;
+  height: 100vh;
+}
+
+.uk-offcanvas.uk-offcanvas-bottom > .uk-offcanvas-bar {
+  width: 100vw;
+  transform: translateX(0) translateY(100%);
+  top: auto;
+}
+
+.uk-offcanvas.uk-offcanvas-bottom.uk-open > .uk-offcanvas-bar {
+  transform: translateX(0) translateY(0);
+}
+</style>
+
+<script>
+import UIkit from "uikit"
+import { uniqueId as _uniqueId } from "lodash"
+
+/**
+ * Create file actions. will be populated by the event payload `filename` and `actions`
+ *
+ */
+export default {
+  name: "oc-file-actions",
+  status: "prototype",
+  release: "1.0.0",
+  props: {
+    /**
+     * The label to display.
+     */
+    label: {
+      type: String,
+      required: false,
+      default: "Open %s in",
+    },
+  },
+  data() {
+    return {
+      filename: "",
+      actions: [],
+      isOpen: false,
+    }
+  },
+  mounted() {
+    this.$root.$on("oc-file-actions:open", file => {
+      this.showActions(file)
+    })
+    UIkit.util.on(`#${this.id}`, "hide", () => {
+      this.closeActions()
+    })
+  },
+  methods: {
+    showActions(file) {
+      this.filename = file.filename
+      this.actions = file.actions
+      this.isOpen = true
+    },
+    closeActions() {
+      this.isOpen = false
+      this.actions = []
+      this.filename = ""
+    },
+    selectAction(action) {
+      action.onClick()
+      this.closeActions()
+    },
+  },
+  computed: {
+    _label() {
+      return this.label.replace(/\%s/, this.filename)
+    },
+    id() {
+      return _uniqueId("oc-fileaction-")
+    },
+  },
+  watch: {
+    isOpen: function(valNew) {
+      if (valNew) {
+        UIkit.offcanvas(`#${this.id}`).show()
+      } else {
+        UIkit.offcanvas(`#${this.id}`).hide()
+      }
+    },
+  },
+}
+</script>
+
+<docs>
+```jsx
+  <template>
+    <div>
+      <oc-button @click="fileOpen(file)" text="click to open actions"/>
+      <oc-file-actions />
+    </div>
+  </template>
+  <script>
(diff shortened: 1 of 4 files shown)


src/patterns/OcFileActions.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <div uk-offcanvas="mode: slide" class="uk-offcanvas uk-offcanvas-bottom">
+    <div class="uk-offcanvas-bar">
+      <span v-text="_label"></span>
+      <ul class="uk-nav">
+        <slot />
```
DeepDiver1975: add documentation to slots

src/patterns/OcFileActions.vue
```diff
@@ -0,0 +1,87 @@
...
+      </ul>
+    </div>
+  </div>
+</template>
+
+<style>
+.uk-offcanvas.uk-offcanvas-bottom {
```
DeepDiver1975: we need to pull this into the less/scss files - maybe use proxy classes .... @felixheidecke

src/elements/OcFileActionItem.vue
```diff
@@ -0,0 +1,70 @@
...
+</template>
+<script>
+/**
+ * Create a file action item.
+ */
+export default {
+  name: "oc-file-action-item",
+  status: "prototype",
+  release: "1.0.0",
+  props: {
```
DeepDiver1975: icon is missing

src/patterns/OcFileActions.vue
```diff
@@ -53,6 +55,15 @@ export default {
       required: false,
       default: "",
     },
+    /**
+     * The actions from which the user can select
+     */
+    items: {
+      type: Array,
```
DeepDiver1975: required: true maybe?

src/patterns/OcFileActions.vue
```diff
@@ -67,18 +78,17 @@ export default {
   <template>
   <div>
     <oc-button uk-toggle="target: #file-actions" text="click to open actions"/>
-    <oc-file-actions id="file-actions"  filename="My file.txt">
-      <oc-file-action-item label="Item 1" @click="onClick" />
-      <oc-file-action-item label="Item 2"/>
-      <oc-file-action-item disabled label="Disabled"/>
-    </oc-file-actions>
+    <oc-file-actions id="file-actions" :items="[{label:'PDF Viewer', icon:'application/pdf', onClick:onPDF},{label:'Text editor', icon:'text/plain', onClick:onTXT}]" filename="My file.txt" />
```
DeepDiver1975: for easier reading you could use a data field for the items - just saying

butonic: @DeepDiver1975 can you point me to an example? I see `model` and `data` being used in the elements but am not sure if thosa are what you mean. I don't see a readable doc for those elements ...

src/patterns/OcFileActions.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <div uk-offcanvas="mode: slide" class="uk-offcanvas uk-offcanvas-bottom oc-file-actions">
+    <div class="uk-offcanvas-bar">
+      <span v-text="_label"></span>
+      <ul class="uk-nav">
+        <template v-for="(item, i) in items">
+          <li uk-toggle="target:.oc-file-actions" :key="i" @click="item.onClick">
```
butonic: @DeepDiver1975 @tomneedham is this the right way to hand over a handler to the vue model?

src/patterns/OcFileActions.vue
butonic: more interesting: how do we pass this in when clicking on a file in the file template?

src/patterns/OcFileActions.vue
DeepDiver1975: https://github.com/owncloud/owncloud-design-system/blob/c856793dd6888602db8cf9477187ab04958b1760/src/patterns/OcAutocomplete.vue#L210

src/patterns/OcFileActions.vue
tomneedham: vue glue code in the files app

DeepDiver1975: props are bound from the parent component

src/patterns/OcFileActions.vue
DeepDiver1975: looks good to me and works ... so :+1:

src/patterns/OcFileActions.vue
```diff
@@ -0,0 +1,142 @@
...
+    },
+  },
+}
+</script>
+
+<docs>
+```jsx
+  <template>
+    <div>
+      <oc-button uk-toggle="target: #file-actions" text="click to open actions"/>
```
DeepDiver1975: we need to get rid of the uk-toggle - see the oc-dialog component - THX

butonic: oc-dialog makes the content of the dialog explicit in template. so it can set up the handlers properly. The fileactions are currently more dynamic, rendering the content based on the action objects. I can revert that and follow the approach the dialog takes, but then using it will no longer be as simple as
`oc-file-actions name="demoActions" :filename="filename" :actions="actions">`.
I mean it isn't anyway ... 

Furthermore, the uikit offcanvas can be closed by clicking outside of it. It fires a hide event, and we can use that to change the internal state accordingly. 

What am I doing wrong? I just went the file actions offcanvas to pop in when a button is clicked. Why do I need to specify anything that has to do with closing the offcanvas in the example at all? that should be self contained, shouldn't it?

DeepDiver1975: how this should work from my understanding:

```html
<button @click='showFileActions=true'/>
<oc-file-actions v-model='showFileActions'/>

DeepDiver1975: but we can address this later as well ... just squash commits and we merge this
this is fine for the first approach

butonic: @DeepDiver1975 I went over this with @felixheidecke. We moved this to a global event system. There is a more beautiful solution for global events, but I consider that outside this PR.

PR #111, opened 2019-03-26
template/login: add first approach

DeepDiver1975:
Bildschirmfoto 2019-04-01 um 10 43 22


felixheidecke:
I guess we need a nice implementation to ship images in a template. Let's add this to a further investigation.

src/templates/Login.vue
```diff
@@ -0,0 +1,41 @@
...
+  name: "Login",
+  status: "deprecated",
+  release: "1.0.0",
+}
+</script>
+<docs>
+```
+<template>
+<div class="oc-login">
+  <div class="oc-login-card uk-position-center">
```
DeepDiver1975: own component: oc-card - possible or do we need a oc-login-card component?

src/templates/Login.vue
```diff
@@ -0,0 +1,41 @@
...
+    <h1 class="oc-login-logo">
+      ownCloud
+    </h1>
+    <div class="oc-login-card-body">
+      <h3 class="oc-login-card-title">
```
DeepDiver1975: oc-card-title

PR #112, opened 2019-03-26
Feature/oc dialog tom

DeepDiver1975:


felixheidecke:
So if we don't want to have all modals sicking around from the beginning, we can simply have a `v-if` condition using `init: false` and set it to `true` once triggered.

DeepDiver1975:
Just Cherry pick my commit

Issue #116, opened 2019-03-27
oc-alert: close handler needed and overlay on top of other components ... 


DeepDiver1975:
1. there is no close handler - we need this from an integration perspective

2. alerts need to be displayed on top of all the other components

tomneedham:
>     1. there is no close handler - we need this from an integration perspective

Implemented in #118 

>     2\. alerts need to be displayed on top of all the other components

Notifications are used for this. Alerts are more persistent, usually present already on the page on load. Appearing above components would not make for great designs. Usually notifications are better for this.

PR #118, opened 2019-03-27
Emit close event if user triggers oc-alert close

tomneedham:
Fixes https://github.com/owncloud/owncloud-design-system/issues/116

Kept the naming to 'close' since we use this with other props.


src/elements/OCAlert.vue
```diff
@@ -43,6 +43,16 @@ export default {
       return classes
     },
   },
+  methods: {
+    onClose(val) {
```
tomneedham: hmm not sure I need the val at all here right?

src/elements/OCAlert.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="oc-alert" :class="_class" uk-alert>
+  <div v-on:hide="onClose" class="oc-alert" :class="_class" uk-alert>
     <a v-if="!noClose" class="uk-alert-close" uk-close></a>
```
DeepDiver1975: maybe emit the event directly from this element and get rid of uk-close?

src/elements/OCAlert.vue
DeepDiver1975: should be the event argument

src/elements/OCAlert.vue
```diff
@@ -43,6 +43,16 @@ export default {
...
     },
   },
+  methods: {
+    onClose(val) {
+      /**
+       * The user closed / hid the alert
+       * @event close
+       * @type {boolean}
+       */
+      this.$emit("close", this.val)
```
DeepDiver1975: not this.val but val aka event - right?

src/elements/OCAlert.vue
tomneedham: In the future we could look at using oc-icon inside here, and emitting a close event. But styling is a mess at the moment trying to do that.

src/elements/OCAlert.vue
tomneedham: yes it is

Issue #94, opened 2019-03-25
component: left app menu


DeepDiver1975:
## screenshots
![Screenshot from 2019-03-25 08-54-19](https://user-images.githubusercontent.com/1005065/54903193-9cfad780-4edb-11e9-84a5-626290aa7378.png)

## elements
- the side bar element which is sliding in and out
- holds a list of menu items (oc-menu variant or oc-table ???)

## behavior
- slide in
- slide out
- sticks to the top menu bar
- does not scroll together with the app content - e.g. file lsit

## events
- ?

## properties
- slot for the inner content

DeepDiver1975:
implemented

PR #121, opened 2019-03-27
introduce oc-applicaiton-menu

scramb:
### Description
Introducing OcLeftSidebar, OcSidebarNavItem, OcSidebarNavSubItem, OcSidebarNavHeader and OcSidebarNavDivider.

### Issues
- refs #94 

### To-Do's
- [ ] Fix text size for nav-items
- [ ] fix appearance in ods e.g. files template
- [ ] ...


src/elements/OcSidebarNavSubItem.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <li :class="[active, 'uk-parent']">
+    <router-link :to="target">
```
tomneedham: Had a quick go at wiring this up in Phoenix. We need to be able to push routes, with params. https://github.com/owncloud/phoenix/blob/master/src/components/Menu.vue#L107 Maybe we just pass in a route object including `name` and `params` elements?

DeepDiver1975: I guess we need that same for the breadcrumbs ....

DeepDiver1975:
@scramb conflicting

src/elements/OcSidebarNavItem.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <li :class="active">
+    <router-link :to="target">
+      <span class="uk-margin-small-right">
+        <oc-icon variation="inverted" :name="icon" />
```
tomneedham: https://getuikit.com/docs/text#vertical-alignment might help to fix alignment of icon - helped for @butonic I think

Issue #125, opened 2019-03-27
oc-text-input: model is not updated


DeepDiver1975:
## steps to reproduce
1. <oc-text-input v-model="inputValue" ..
2. enter text into the input fiels
3. see the data element inputValue not being updated

DeepDiver1975:
+ please add a method to put focus on this element

DeepDiver1975:
fixed with #131

PR #131, opened 2019-03-28
Bugfix/125

DeepDiver1975:
- implement v-model behavior properly
- apply https://vuejs.org/v2/style-guide/#Private-property-names-essential
- add some more example code
- add public method focus() to set the focus

### DIFF
--- a/src/elements/OcTextInput.vue
+++ b/src/elements/OcTextInput.vue
@@ -5,9 +5,10 @@
     :value="value"
     :placeholder="placeholder"
     :aria-label="placeholder"
-    @input="onInput($event.target.value)"
-    @focus="onFocus($event.target.value)"
-    @keydown="onKeyDown($event)"
+    ref="input"
+    @input="$_ocTextInput_onInput($event.target.value)"
+    @focus="$_ocTextInput_onFocus($event.target.value)"
+    @keydown="$_ocTextInput_onKeyDown($event)"
   />
 </template>
 
@@ -35,6 +36,7 @@ export default {
     },
     /**
      * Text value of the form input field.
+     * @model
      */
     value: {
       type: String,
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
fixes #125

src/elements/OcTextInput.vue
```diff
@@ -5,19 +5,24 @@
...
-    @focus="onFocus($event.target.value)"
+    ref="input"
+    @input="$_ocTextInput_onInput($event.target.value)"
+    @focus="$_ocTextInput_onFocus($event.target.value)"
+    @keydown="$_ocTextInput_onKeyDown($event)"
   />
 </template>
 
 <script>
+import OcButton from "./OcButton"
```
butonic: AFAICT the oc-text-input does not need an ocbutton. only your doc bloc does

src/elements/OcTextInput.vue
```diff
@@ -5,19 +5,24 @@
...
+
 /**
  * Form Inputs are used to allow users to provide text input when the expected
  * input is short. Form Input has a range of options and supports several text
  * formats including numbers. For longer input, use the form `Textarea` element.
  */
 export default {
   name: "oc-text-input",
+  components: { OcButton },
```
butonic: no, not really

DeepDiver1975: I know - could not find an alternative approach
use a plain html button instead?

butonic: um I don't think you need to reference it:
```diff
diff --git a/src/elements/OcTextInput.vue b/src/elements/OcTextInput.vue
index 8c85fc6..768e3b6 100644
--- a/src/elements/OcTextInput.vue
+++ b/src/elements/OcTextInput.vue
@@ -110,7 +110,7 @@ export default {
             <h3 class="uk-heading-divider">
                 Interactions
             </h3>
-            <button @click="_focus">Focus input below</button>
+            <oc-button text="Focus input below" @click="_focus" />
             <oc-text-input placeholder="Will you focus on me?" ref="inputForFocus"/>
         </section>
     </template>
```

works just fine for the button

PR #149, opened 2019-04-01
add initial app layout

butonic:
initial app layout with optional left and right bar
- used it to add right sidebar in files template
- also needed adding a click event to the table row

### DIFF
--- a/src/elements/_OcTableRow.vue
+++ b/src/elements/_OcTableRow.vue
@@ -1,5 +1,5 @@
 <template>
-  <tr>
+  <tr @click="onClick">
     <slot />
   </tr>
 </template>
@@ -8,5 +8,15 @@ export default {
   name: "oc-table-row",
   status: "review",
   release: "1.0.0",
+  methods: {
+    onClick(val) {
+      /**
+       * Click event
+       * @event click
+       * @type {event}
+       */
+      this.$emit("click", val)
+    },
+  },
 }
 </script>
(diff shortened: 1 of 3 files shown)


tomneedham:
Talking with @felixheidecke (https://talk.owncloud.com/channel/phoenix?msg=LtiDXhP8Rs5CyNQGK) - this might not be the right direction. It doesn't add a huge amount of functionality, and instead it restrict flexibility in the layout depending on the app (think: responsiveness, other layouts).

Instead, this is easily implemented per app as a simple div.uk-grid, and v-if to toggle the sidebars.

DeepDiver1975:
> Talking with @felixheidecke (https://talk.owncloud.com/channel/phoenix?msg=LtiDXhP8Rs5CyNQGK) - this might not be the right direction. It doesn't add a huge amount of functionality, and instead it restrict flexibility in the layout depending on the app (think: responsiveness, other layouts).

To me the absolute opposite is the case: we can add respsoniveness (and other future behavior) to the component in ODS and ALL places where the component is used will benefit from the improvments.

butonic:
This really is a matter of priorities. Maybe we can reuse this component. If so, changes will affect all places that use it. That might be desired. It might actually NOT be desired. Maybe we cannot reuse it then implementing it was a waste of time. We'll never know, unless we iterate on this and then drill down on the requirements.

We already kicked out vuetify because it doesn't fit properly. Maybe we need to test this and then iterate again.

PR #171, opened 2019-04-03
Added slots in TabItem, Breadcrumb and SidebarNavItem

LukasHirt:
Added slots in TabItem, Breadcrumb and SidebarNavItem

### DIFF
--- a/src/elements/OcBreadcrumb.vue
+++ b/src/elements/OcBreadcrumb.vue
@@ -44,13 +44,7 @@ export default {
             {text:'Folder',to:{path:'folder'}},
             {text:'Subfolder', onClick:() => alert('Breadcrumb clicked!')}]
         }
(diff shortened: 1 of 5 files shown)


LukasHirt:
Added some changes into the description of the text and title props about how they are used with slots. Not sure if you meant it like this so let me know if I got this one wrong.

DeepDiver1975:
> Added some changes into the description of the text and title props about how they are used with slots. Not sure if you meant it like this so let me know if I got this one wrong.

I was talking about the ```<docs>``` section at the end of the file. we shall use the components down there with the slots ...

LukasHirt:
I see. Should be fixed now.

DeepDiver1975:
> I see. Should be fixed now.

better - thx

PR #174, opened 2019-04-03
Render a component/template in fullscreen

DeepDiver1975:
## Description

this add a button with will render the component in full screen - this helps in designing the component for different screen sizes

Screenshot from 2019-04-03 18-20-44

## ToDo
- [x] the background needs to be set properly

### DIFF
--- a/docs/components/Preview.js
+++ b/docs/components/Preview.js
@@ -43,6 +43,15 @@ export default previewComponent => {
         this.$el.localName +
         ">"
 
+      let button = document.createElement("button")
+      button.innerHTML = "Display component in fullscreen"
+      button.onclick = function() {
+        button.parentNode.requestFullscreen()
+        button.remove()
+        return false
+      }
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
now white ....

Screenshot from 2019-04-03 18-41-46

michaelstingl:
This is a great idea! Something we should contribute upstream?

DeepDiver1975:
> After entering into fullscreen the button disappears. If I leave it then I can't go back into it without refreshing the page

@LukasHirt take care of - please review - THX

LukasHirt:
> > After entering into fullscreen the button disappears. If I leave it then I can't go back into it without refreshing the page
> 
> @LukasHirt take care of - please review - THX

Yes, works fine now. Also I wasn't able to scroll in the full screen. But that was fixed with just adding overflow: auto, so I fixed it already.

PR #183, opened 2019-04-04
Add npm publish steps

DeepDiver1975:
This will publish ods to npmjs repo. Currently each master branch push will be built and pushed there ....

Support for tags will follow ....

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,6 +2,13 @@ branches:
   - master
 
 pipeline:
+  increment-version:
+    image: node
+    commands:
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -14,9 +21,15 @@ pipeline:
...
     secrets: [github_username, github_password]
     pages_directory: dist/docs
     when:
       event: [push]
+
+  publish-system:
+    image: plugins/npm:1
+    secrets: [npm_username, npm_email, npm_token]
+    when:
+      event: [push]
```
tboerger: Also on tag, right?

DeepDiver1975: yes - I'll change this once we start tagging - I need to create a different version then in package.json

Issue #140, opened 2019-04-01
oc-top-bar: sticky


DeepDiver1975:
the topbars need the capability to stick to the browser window.

cauton: this also applies in case where we have two top bars ....

PR #191, opened 2019-04-05
Add stickyness to topbar

felixheidecke:
Restructure topbar to prevent blank items

Fixes: https://github.com/owncloud/owncloud-design-system/issues/140

Also: 
Bildschirmfoto 2019-04-05 um 13 33 22


src/elements/OcBreadcrumb.vue
```diff
@@ -26,6 +29,17 @@ export default {
       default: [],
       required: true,
     },
+    firstIsHome: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  methods: {
+    goHome() {
```
DeepDiver1975: Please follow https://vuejs.org/v2/style-guide/#Private-property-names-essential

```$_ocBreadCrumb_goHome

File: src/elements/OcBreadcrumb.vue [hunk 2]
```diff
@@ -26,6 +29,17 @@ export default {
...
+      // We need the router here
+      //
+      // this.router.push('/')
```
DeepDiver1975: simply emit an event here?

src/patterns/_OcNavbarItem.vue
```diff
@@ -0,0 +1,21 @@
...
+  status: "review",
+  release: "1.0.0",
+  props: {
+    position: {
+      type: String,
+      default: "left",
+    },
+  },
+  computed: {
+    _class() {
```
DeepDiver1975: naming - see above https://vuejs.org/v2/style-guide/#Private-property-names-essential

src/templates/Topbar.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <oc-navbar tag="header" class="oc-topbar">
+    <oc-navbar-item position="left">
+      <button class="oc-topbar-menu-burger" aria-label="Files">
```
DeepDiver1975: oc-button ?

src/templates/Topbar.vue
```diff
@@ -0,0 +1,39 @@
+<template>
```
DeepDiver1975: shall we use the template as is in phoenix? or shall we copy over the template?

src/templates/Topbar.vue
felixheidecke: I needed a "blank" button tag with no styling here.

src/templates/Topbar.vue
felixheidecke: I recommend using as is. There will be no other places where this should or could be used.

DeepDiver1975: ok - so basically copying over the template section - right?

src/elements/OcBreadcrumb.vue
felixheidecke: I could also leave it blank as we wouldn't need to catch that event anywhere else but within itself. Agree?

DeepDiver1975: how would a user of the component then get the event? Currently this is not usable

src/pages/Files.vue
```diff
@@ -88,59 +95,54 @@ export default {
...
-        <oc-breadcrumb :items="[{text:'Home',to:'/'},{text:'Folder',to:{path:'folder'}},{text:'Subfolder'}]"></oc-breadcrumb>
-      </template>
-        <template slot="title">
-            <div class="uk-navbar-item">
-                <oc-search-bar label="Search"></oc-search-bar>
+    <oc-app-layout :rightHidden="rightHidden">
+
+      <template slot="center">
+        <section class="uk-background-muted uk-padding-small">
+          <div uk-grid class="uk-grid-small uk-flex-middle">
```
DeepDiver1975: the files top bar is no longer a navbar component? It should surly be - right?

src/pages/Files.vue
```diff
@@ -88,59 +95,54 @@ export default {
...
-                      </oc-menu-item>
-                      <oc-menu-item>
-                        Hidden files<oc-checkbox></oc-checkbox>
-                      </oc-menu-item>
-                      <oc-menu-item>
-                        Files by name<oc-text-input></oc-text-input>
-                      </oc-menu-item>
-                    </template>
-                </oc-menu>
+            <div class="uk-visible@l">
```
DeepDiver1975: the filter menu will be a component of it's own - see the proposal if @LukasHirt in #197

src/pages/Files.vue
felixheidecke: It requires a couple more sub-components. Grid, Grid Items and then the app-bar. These should become a part of a different PR is suppose. The Grid-component may become pretty complex.

DeepDiver1975: well - this sub-topbar is the more important component because it is reused in the tree apps we have as of now. We need to address this right in here as well. 

Otherwise this will totally break phoenix apart.

src/pages/Files.vue
felixheidecke: I removed it so we can plug in the component later.

DeepDiver1975:
Thx . I will Squash commits and merge later

src/styles/theme/oc-topbar.scss
```diff
@@ -1,3 +1,59 @@
-.oc-topbar.uk-light .uk-navbar-item {
```
DeepDiver1975: looks like this change is killing the secondary color of oc-topbar ....

PR #194, opened 2019-04-08
Emit change event when checkbox model changes

tomneedham:
Fix https://github.com/owncloud/owncloud-design-system/issues/134 based on https://stackoverflow.com/a/51432277/6795549


DeepDiver1975:
I pushed some update on the docs section - change event works - but data binding is broken ....

PR #196, opened 2019-04-08
Added translation

LukasHirt:
## Description
Added translation into the menu link

### DIFF
--- a/src/elements/OcMenuItem.vue
+++ b/src/elements/OcMenuItem.vue
@@ -1,16 +1,13 @@
 <template>
   <li @click="onClick">
     <a href="#">
-      <!-- @slot This default slot is set as content & will override icon -->
-      <slot name="default">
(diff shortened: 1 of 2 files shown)


src/elements/OcMenuItem.vue
```diff
@@ -5,10 +5,10 @@
       <slot name="default">
         <div v-if="icon" class="uk-flex uk-flex-middle">
           <oc-icon :name="icon" class="uk-margin-small-right" />
-          {{ content }}
+          <template>{{ content }}</template>
         </div>
         <template v-else>
-          {{ content }}
+          <template v-translate>{{ content }}</template>
```
DeepDiver1975: this will not work .... v-translate has to remain in phoenix.

LukasHirt: I removed the translations and edited slots so now it is able to use translations via the slot in Phoenix and still preserve the classes for alignment when the item has an icon.

PR #197, opened 2019-04-08
Created filterMenuItem component

LukasHirt:
## Description
Created a new menu item component for filter menu

## Motivation
By creating own component we don't need to add every time checkboxes to every item and also the item doesn't need to be wrapped in a link.

### DIFF
--- a/src/elements/OcFilterMenuItem.vue
+++ b/src/elements/OcFilterMenuItem.vue
@@ -1,0 +1,113 @@
+<template>
+  <li>
+    <slot name="searchBar">
+      <div class="uk-flex uk-flex-middle">
+        <oc-checkbox
+          :id="name"
+          :value="value"
+          class="uk-margin-small-right"
+          @change="$_ocCheckbox_onChange($event)"
+        />
+        <label :for="name">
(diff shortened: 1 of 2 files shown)


src/elements/OcFilterMenuItem.vue
```diff
@@ -0,0 +1,68 @@
...
+  </template>
+  <script>
+      export default {
+          data: () => {
+              return {
+                  filterQuery: ''
+              }
+          },
+          methods: {
+              onFilter(val) {
```
DeepDiver1975: please respect naming $_ocFilterMenuItem_onFilter

src/elements/OcFilterMenuItem.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <li>
+    <slot name="default">
+      <div class="uk-flex uk-flex-middle">
+        <oc-checkbox class="uk-margin-small-right" />
```
DeepDiver1975: - [ ] event when changing the checkbox needs to be emitted
- [ ] when clicking on the label usually a change on the corresponding checkbox is triggered

PR #198, opened 2019-04-08
Created upload-menu component

LukasHirt:
## Description
Created the upload menu component.

## Motivation
Old menu didn't use UIkit. With UIkit it looks a little bit better (still not perfect but I'll return to this to style lit a little bit more) and was hard to orientate in.

### DIFF
--- a/src/patterns/OcUploadMenu.vue
+++ b/src/patterns/OcUploadMenu.vue
@@ -1,0 +1,75 @@
+<template>
+  <ul class="uk-list uk-margin-remove">
+    <li v-for="(item, index) in items" :key="index">
+      <slot name="default">
+        <div class="uk-flex">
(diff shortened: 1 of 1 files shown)


src/patterns/OcUploadMenu.vue
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Create a upload menu item component
+ */
+export default {
+  name: "oc-upload-menu",
+  status: "review",
+  release: "0.0.1",
+  props: {
+    /**
+     * Array with items to display in menu
```
DeepDiver1975: - [ ] describe properties of the items: name, size and progres

LukasHirt: Done

PR #212, opened 2019-04-10
Created menu list item component

LukasHirt:
## Motivation
In some cases, we need menu items that looks the same as link items but would be better if they have no link element (upload, for example, doesn't work if it is wrapped in a link).

### DIFF
--- a/src/elements/OcMenuItem.vue
+++ b/src/elements/OcMenuItem.vue
@@ -1,14 +1,12 @@
 <template>
-  <li @click="onClick">
-    <a href="#">
-      <div v-if="icon" class="uk-flex uk-flex-middle">
-        <oc-icon :name="icon" class="uk-margin-small-right" />
-        <slot name="default" />
-      </div>
-      <template v-else>
-        <slot name="default" />
-      </template>
-    </a>
+  <li class="oc-menu-item" @click="$_ocMenuItem_onClick">
+    <div v-if="icon" class="uk-flex uk-flex-middle">
+      <oc-icon :name="icon" class="uk-margin-small-right" />
+      <slot name="default" />
+    </div>
+    <template v-else>
+      <slot name="default" />
+    </template>
   </li>
 </template>
 <script>
@@ -30,7 +28,7 @@ export default {
     },
   },
   methods: {
-    onClick(val) {
+    $_ocMenuItem_onClick(val) {
       /**
        * Click event
        * @event click
@@ -47,7 +45,7 @@ export default {
 <template>
     <oc-menu buttonText="open oc-menu">
       <template slot="subnav">
-        <oc-menu-item icon="create_new_folder">Item with icon</oc-menu-item>
+        <oc-menu-item icon="create_new_folder" @click="onClick">Item with icon</oc-menu-item>
         <oc-menu-item>Item without icon</oc-menu-item>
       </template>
     </oc-menu>
@@ -55,8 +53,8 @@ export default {
   <script>
     export default {
       methods: {
-        onClick(val) {
-          alert(val)
+        onClick() {
+          alert("Hello world!")
         }
       }
     }
--- a/src/patterns/OcMenu.vue
+++ b/src/patterns/OcMenu.vue
@@ -6,7 +6,7 @@
     <div uk-dropdown="mode: click, offset: 0">
       <!-- @slot The default slot for inner content  -->
       <slot name="default"> </slot>
-      <ul class="uk-nav uk-dropdown-nav uk-nav-default">
+      <ul class="uk-list uk-margin-remove-bottom">
         <!-- @slot This slot receives oc-menu-items as content -->
         <slot name="subnav" />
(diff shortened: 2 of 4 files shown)


LukasHirt:
> Why not remove the a tag from OcMenuItem? I see no need in a second component which looks 99% identical. If we still need both variants please add a property to OcMenuItem.

I removed a tag and added class to ensure the same spacing and, pointer cursor and colour (related to https://github.com/owncloud/phoenix/pull/905). Also, I rebased the branch so the previous commit should be gone.

DeepDiver1975:
@felixheidecke can you please review the scss magic in here? THX

src/elements/OcMenuItem.vue
```diff
@@ -29,6 +27,12 @@ export default {
       default: "",
     },
   },
+  computed: {
+    $_ocMenuItem_menuItemClass() {
```
DeepDiver1975: no need to apply a constant via a computed property - right?

src/elements/OcMenuItem.vue
```diff
@@ -29,6 +27,12 @@ export default {
...
+      let classes = ["oc-menu-item"]
+      return classes
+    },
+  },
   methods: {
     onClick(val) {
```
DeepDiver1975: please apply naming convention to onClick - THX

src/elements/OcMenuItem.vue
LukasHirt: Right. Fixed

src/elements/OcMenuItem.vue
LukasHirt: Fixed

src/tokens/color.yml
```diff
@@ -23,6 +23,12 @@ props:
   link-hover-color:
     value: "#0066ff"
     category: "Font-colors"
+  menu-item-color:
```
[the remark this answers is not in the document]
LukasHirt: I don't think it's such a good approach to have the same colour for a link in navigation and link in the content. What is the reason behind not introducing new colour? I mean I'm almost 100 % sure we would need to introduce this colour anyway in the future.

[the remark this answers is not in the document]
felixheidecke: Currently, all UI component colors are based on the color-set we have here: https://owncloud.github.io/owncloud-design-system/#/Design%20Tokens

If these base colors are changed, the rest of the components will adapt to those basic colors, making it easy to change color pallet. I'd rather you fall back to the default colors for now that the base component supplies.

If a theme designer decides to change colors, he can do so by using either component hooks, change component base variables or individualize using the class-proxy approach.

UIKit uses more than 500 variables that draw from the global variable-set that is exposed in the design tokes. They should be customized on an **optional** basis.

Understandable? 🤪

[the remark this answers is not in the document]
LukasHirt: I get how it works. I'm just trying to say that nav item colour seems to me like a pretty basic colour to have in the design tokens. But okay, I'll get it out of tokens.

src/styles/theme/oc-menu-item.scss
```diff
@@ -0,0 +1,22 @@
+//
+// Component: OC MenuItem
+//
+// ========================================================================
+.oc-menu-item {
+  padding: $menu-item-padding;
```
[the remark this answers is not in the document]
LukasHirt: .uk-list use margin instead of padding which I want to use at links in menu. Padding is used in .uk-nav but it adds it to <a> instead of <li>

src/elements/OcMenuItem.vue
```diff
@@ -1,14 +1,12 @@
...
-      <div v-if="icon" class="uk-flex uk-flex-middle">
-        <oc-icon :name="icon" class="uk-margin-small-right" />
-        <slot name="default" />
-      </div>
-      <template v-else>
-        <slot name="default" />
-      </template>
-    </a>
+  <li class="oc-menu-item" @click="$_ocMenuItem_onClick">
+    <div v-if="icon" class="uk-flex uk-flex-middle">
```
felixheidecke: We need to have the icon wrapped in the anchor tag, so it becomes clickable as well.

src/patterns/OcMenu.vue
```diff
@@ -6,7 +6,7 @@
     <div uk-dropdown="mode: click, offset: 0">
       <!-- @slot The default slot for inner content  -->
       <slot name="default"> </slot>
-      <ul class="uk-nav uk-dropdown-nav uk-nav-default">
+      <ul class="uk-list uk-margin-remove-bottom">
```
felixheidecke: The design of the component is "too narrow" in the sense of flexibility. Better to create a separate `oc-nav` component and so it can be mixed in in the slot where needed.

DeepDiver1975: No ul should be in here. Just ocmenuitems

DeepDiver1975:
superseded by #226

PR #226, opened 2019-04-16
Introduce OCDrop Element

felixheidecke:
Introduced the `<oc-drop />` Element that allows devs to open overlaying containers for little menus for example.

Bildschirmfoto 2019-04-16 um 18 04 53
Bildschirmfoto 2019-04-16 um 18 05 01

### Todo

- [x] Create oc-nav
- [x] Extend oc-nav-item
- [x] Use the above in the oc-drop example
- [x] Fix/enhance documentation


felixheidecke:
How it difference to oc-menu:

- Can be triggered from outside the component using id, class or attribute selector
- Boundary defaults to the trigger, but it can be placed elsewhere
- The default slot provides a uk-card wrapper for easy filling with "simple" contents
- Supplies an extra, unwrapped slot for any components already bringing their own styles _(can be seen in the example)_
- Can use the whole range of properties uk-drop provides via the options attribute

src/elements/OcNavItem.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <li @click="onClick">
-    <a href="#">
+  <li :class="{ 'uk-active': active }">
+    <a href="#" @click.prevent="onClick">
```
DeepDiver1975: Naming. Please adjust

src/elements/OcNavItem.vue
```diff
@@ -45,21 +55,20 @@ export default {
...
-      </template>
-    </oc-menu>
-  </template>
-  <script>
-    export default {
-      methods: {
-        onClick(val) {
-          alert(val)
-        }
+  <oc-nav>
```
DeepDiver1975: Can you please Model the Filter menu as example Here as Well? Thx

src/elements/OCDrop.vue
```diff
@@ -0,0 +1,99 @@
+<template>
+  <div :uk-drop="$_ocDrop_props">
+    <div v-if="$slots.default" class="uk-card uk-card-default uk-card-small uk-card-body">
+      <slot />
+    </div>
+    <slot name="special" />
```
DeepDiver1975: Else is Missing. Or is IT intended that both Slots can be rendered

src/elements/OcNavItem.vue
felixheidecke: I added the filter menu to oc-drop

src/elements/OCDrop.vue
```diff
@@ -0,0 +1,117 @@
...
+    },
+    /**
+     * CSS selector for the element to be used as toggle. By default, the preceding element is used.
+     **/
+    toggle: {
+      type: String,
+      default: "- *",
+    },
+    /**
+     * Comma separated list of drop trigger behaviour modes
```
DeepDiver1975: please add allowed values to the docs

src/elements/OCDrop.vue
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Position any element in relation to another element.
+ */
+export default {
+  name: "oc-drop",
+  status: "review",
+  release: "1.0.0",
+  props: {
+    /**
+     * CSS selector of the element to maintain the drop's visibility.
```
DeepDiver1975: I don't understand what this is for. Needs better explanation or maybe even an example. THX

src/elements/OCDrop.vue
```diff
@@ -0,0 +1,117 @@
...
+     **/
+    mode: {
+      type: String,
+      default: "click",
+      validator: value => {
+        return value.match(/(click|hover)/)
+      },
+    },
+    /**
+     * More options are available
```
DeepDiver1975: needs more explanation - maybe add link to uikit?

src/elements/OCDrop.vue
DeepDiver1975: taken care off

src/elements/OCDrop.vue
DeepDiver1975: take care off

src/elements/OCDrop.vue
DeepDiver1975: taken care off

DeepDiver1975:
THX @felixheidecke - will merge asap

PR #245, opened 2019-04-25
Added slots

### LINKED ISSUE #200: No user visible strings shall live inside ODS
Any user facing string/text has to be translated.
Translations are considered to be in the responsibility of the consuming project.

- Any text has to be passed in via properties/slots
- No translation function is to be used within ODS

LukasHirt:
## Description
Added slots instead of text props.

## Related issue
- Fixes #200

### DIFF
--- a/src/elements/OcButton.vue
+++ b/src/elements/OcButton.vue
@@ -4,15 +4,14 @@
     :href="href"
     :to="to"
     :type="submit"
-    :text="text"
     :aria-label="ariaLabel"
     :class="$_ocButton_buttonClass"
     @click="onClick"
   >
-    <oc-icon v-if="icon" :class="$_ocButton_iconClass" :name="icon"></oc-icon>
-    <slot>
-      <span v-if="text" v-text="text" />
-    </slot>
+    <oc-icon v-if="icon" :class="$_ocButton_iconClass" :name="icon" />
+    <span v-if="$slots.default">
+      <slot name="default" />
+    </span>
   </component>
 </template>
 
@@ -99,17 +98,12 @@ export default {
         return value.match(/(default|primary|secondary)/)
       },
     },
-    /**
-     * Text do be displayed in the button
-     */
-    text: {
-      type: String,
-      default: null,
-    },
   },
   computed: {
     $_ocButton_iconClass() {
-      return this.text ? "uk-position-small uk-position-center-left" : "uk-position-center"
+      return this.$slots.default
+        ? "uk-position-small uk-position-center-left"
+        : "uk-position-center"
     },
     $_ocButton_buttonClass() {
       let classes = ["oc-button"]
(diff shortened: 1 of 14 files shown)


src/elements/_OcAccordionItem.vue
```diff
@@ -1,11 +1,13 @@
 <template>
   <li>
     <a class="uk-accordion-title" href="#">
-      {{ title }}
-      <span v-if="subtitle !== undefined" class="uk-text-meta uk-margin-left">{{ subtitle }}</span>
+      <slot name="title" />
+      <span v-if="$slots.subtitle" class="uk-text-meta uk-margin-left"
+        ><slot name="subtitle"
```
DeepDiver1975: line breaks looks a bit strange ...

src/elements/_OcAccordionItem.vue
```diff
@@ -14,28 +16,15 @@ export default {
...
-      type: String,
-      required: false,
-    },
-  },
 }
 </script>
 <docs>
 ```jsx
 <oc-accordion>
   <oc-accordion-item title="Public link share">
```
DeepDiver1975: remove title then?

src/elements/_OcAccordionItem.vue
```diff
@@ -14,28 +16,15 @@ export default {
...
+    <template slot="title">
```
DeepDiver1975: maybe use
```
<div slot=title>Public link share<div>

File: src/elements/_OcSidebarNavItem.vue [hunk 4]
```diff
@@ -38,11 +34,6 @@ export default {
       required: false,
       default: "",
     },
-    text: {
```
DeepDiver1975: please research on vuejs v-text - there was some special handling/support .... the text can be entered via prop or in a slot ..... if I recall correctly

src/elements/OcButton.vue
```diff
@@ -4,15 +4,14 @@
...
     :aria-label="ariaLabel"
     :class="$_ocButton_buttonClass"
     @click="onClick"
   >
-    <oc-icon v-if="icon" :class="$_ocButton_iconClass" :name="icon"></oc-icon>
-    <slot>
-      <span v-if="text" v-text="text" />
-    </slot>
+    <oc-icon v-if="icon" :class="$_ocButton_iconClass" :name="icon" />
+    <span>
```
felixheidecke: This `<span>` will also show up in icon-only buttons and add an unwanted space on the right.
Let's `v-if` this.

src/elements/_OcAccordionItem.vue
LukasHirt: Fixed

src/elements/_OcAccordionItem.vue
LukasHirt: Done

src/elements/_OcAccordionItem.vue
LukasHirt: Fixed

src/elements/OcButton.vue
LukasHirt: Done

src/elements/OcButton.vue
```diff
@@ -4,15 +4,14 @@
     :href="href"
     :to="to"
     :type="submit"
-    :text="text"
```
DeepDiver1975: let's keep v-text here - see https://vuejs.org/v2/api/#v-text .... maybe somthing to consider in different places as well ....

LukasHirt: I'm not sure I understand how is this supposed to help. If you include it with v-text then you need to write again getText - right? Which in the end I think is the same amount of work as to write it into its own slot.

DeepDiver1975: The textContent can also be taken from the inner node I I understand correctly. That is my point

src/elements/OcButton.vue
```diff
@@ -4,15 +4,14 @@
...
     :aria-label="ariaLabel"
     :class="$_ocButton_buttonClass"
     @click="onClick"
   >
-    <oc-icon v-if="icon" :class="$_ocButton_iconClass" :name="icon"></oc-icon>
-    <slot>
-      <span v-if="text" v-text="text" />
-    </slot>
+    <oc-icon v-if="icon" :class="$_ocButton_iconClass" :name="icon" />
+    <span v-if="$slots.default">
```
DeepDiver1975: This change is okay - I did missunderstand this all ... sorry.

src/elements/_OcSidebarNavItem.vue
```diff
@@ -6,14 +6,10 @@
...
-      <slot name="text">
-        <span>
-          {{ text }}
-        </span>
-      </slot>
-    </component>
-    <ul class="uk-nav-sub" v-if="$_ocSidebarNavItem_hasDefaultSlot">
       <slot name="default" />
+    </component>
+    <ul class="uk-nav-sub" v-if="$slots.subnav">
```
DeepDiver1975: I'd prefer to keep the sub navs in the default slot.
any text can go into a text slot

LukasHirt: Since text will be implemented more often than subnav I think it would mean more work than having it as a default slot.

PR #253, opened 2019-04-29
Loading string can be set via a property + changing the model is now …

DeepDiver1975:
…triggering an event


felixheidecke:
I just merged this, so it won't collide with my update with complex contents.

PR #261, opened 2019-04-30
Complex autocomplete

felixheidecke:
Allows for custom suggestion rendering and complex objects.

type
select

### DIFF
--- a/src/patterns/OcAutocomplete.vue
+++ b/src/patterns/OcAutocomplete.vue
@@ -7,16 +7,16 @@
       :disabled="disabled"
       @keydown.up.prevent="highlighted--"
       @keydown.down.prevent="highlighted++"
-      @keydown.enter="selectSuggestion"
+      @keydown.enter="$_ocAutocomplete_selectSuggestion"
     />
-    <div hidden :id="_boundryId" />
+    <div hidden :id="$_ocAutocomplete_boundryId" />
     <div
       class="oc-autocomplete-dropdown"
-      :uk-drop="'mode:click;delay-hide:0;toggle:#' + _boundryId"
-      :id="_dropdownId"
+      :uk-drop="'mode:click;delay-hide:0;toggle:#' + $_ocAutocomplete_boundryId"
+      :id="$_ocAutocomplete_dropdownId"
     >
       <ul class="oc-autocomplete-suggestion-list">
-        <template v-for="(item, i) in matchesShown">
+        <template v-for="(item, i) in $_ocAutocomplete_matchesShown">
           <li
             :class="[
               'oc-autocomplete-suggestion',
@@ -24,7 +24,7 @@
             ]"
             :key="i"
             @mouseenter="highlighted = i"
-            @click="selectSuggestion"
+            @click="$_ocAutocomplete_selectSuggestion"
           >
             <slot name="item" v-bind:item="item">
               <!-- Fallback content -->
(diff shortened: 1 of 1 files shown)


felixheidecke:
I reverted back to the `findBy` property as injecting your own method messed up the rest of the code since I didn't find a way to calculate with the result in the component.

DeepDiver1975:
Nice. THX. I can help Out with the Filter property in thursday.

src/patterns/OcAutocomplete.vue
```diff
@@ -7,33 +7,37 @@
...
         </template>
         <li
-          v-if="matchesOverflowing > 0 && !itemsLoading"
+          v-if="$_ocAutocomplete_matchesOverflowing > 0 && !itemsLoading"
           class="oc-autocomplete-suggestion-overflow"
         >
-          {{ matchesOverflowing }}
-          <span v-if="matchesOverflowing === 1">more result</span>
+          {{ $_ocAutocomplete_matchesOverflowing }}
+          <span v-if="$_ocAutocomplete_matchesOverflowing === 1">more result</span>
```
DeepDiver1975: we need a property where we can set this user facing string from outside - because of translations ...

src/patterns/OcAutocomplete.vue
```diff
@@ -114,10 +118,10 @@ export default {
       default: 5,
     },
     /**
-     * The selected element
+     * Select what values an object is to be found by
      */
-    value: {
```
DeepDiver1975: why was value removed? this is the model which shall hold the selected element.
The parent component needs this to get informed about selection.

please bring it back. THX

src/patterns/OcAutocomplete.vue
```diff
@@ -157,39 +172,41 @@ export default {
...
        */
       this.$emit("update:input", value)
     },
-    selectSuggestion() {
-      if (this.matchesShown[this.highlighted]) {
-        this.input = this.matchesShown[this.highlighted]
-        this.$emit("input", this.input)
+    $_ocAutocomplete_selectSuggestion() {
+      if (this.$_ocAutocomplete_matchesShown[this.highlighted]) {
+        this.$emit("select", this.$_ocAutocomplete_matchesShown[this.highlighted])
```
DeepDiver1975: please emit 'input' as before. THX

src/patterns/OcAutocomplete.vue
```diff
@@ -157,39 +172,41 @@ export default {
...
+        this.input = ""
+        this._dropdown.hide()
```
DeepDiver1975: _dropdown unknown ...

DeepDiver1975:
> I reverted back to the `findBy` property as injecting your own method messed up the rest of the code since I didn't find a way to calculate with the result in the component.

https://github.com/owncloud/owncloud-design-system/pull/263

src/patterns/OcAutocomplete.vue
felixheidecke: We could have one `text` property that expects an object like:

`{ loading : 'Waiting for results', xMoreResults : '%s more results', oneMoreResult : '1 more result' }`

src/patterns/OcAutocomplete.vue
felixheidecke: The `input` value does not reflect the `output` value, as we won't be able to display an object inside the `<input />` element. There is the `@select` that returns the value once selected. I still can wire up the input, should I?

src/patterns/OcAutocomplete.vue
felixheidecke: As mentioned `input != output`

felixheidecke: I can emit input "as well"

PR #264, opened 2019-05-02
Created oc-file-actions class and changed colour of actions bar

LukasHirt:
## Description
Created oc-file-actions class and moved styles from the components into it and changed colour of actions bar. It's still not perfect colour but I would say it's the most appropriate for that bar from what we have in design tokens. Also it's just temporary so as long as it makes the links a little bit more visible I think it's fine for the time being.


DeepDiver1975:
Some icons are not visible because of coloring ...

Screenshot from 2019-05-03 08-17-22

LukasHirt:
> Some icons are not visible because of coloring ...
> 
> Screenshot from 2019-05-03 08-17-22

Should be fixed now.

PR #323, opened 2019-06-04
New App-Bar

felixheidecke:
* Add padding and background to app-bar

Only works in conjunction with https://github.com/owncloud/phoenix/pull/1221

### DIFF
--- a/src/styles/_owncloud.scss
+++ b/src/styles/_owncloud.scss
@@ -34,6 +34,8 @@
 @import "theme/oc-user-menu";
 @import "theme/oc-dropzone";
 @import "theme/oc-file-actions";
+@import "theme/oc-app-bar";
+@import "theme/oc-scroll";
 
 // 4. Import UIkit.
 @import "../../node_modules/uikit/src/scss/uikit-theme.scss";
--- a/src/styles/theme/main.scss
+++ b/src/styles/theme/main.scss
@@ -1,3 +1,8 @@
+body {
+  height: 100vh;
+  overflow: hidden;
+}
+
 #oc-content {
   height: 100vh;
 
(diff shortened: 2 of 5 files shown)


src/elements/OcBreadcrumb.vue
```diff
@@ -25,7 +25,7 @@ export default {
   release: "1.0.0",
   props: {
     items: {
-      type: Array,
+      type: [Boolean, Array],
```
DeepDiver1975: Is that reasonable? What Happens If Set to true? What when Set to false

felixheidecke: Problem was, that `activeRoute` returns `false` in the home directory. Since `home` is now part of the breadcrumb, a type warning pops up.

DeepDiver1975: but the correct fix is in phoenix to return [] instead of false

felixheidecke: I'm wondering if root should be the first entry in the array so we can navigate to it properly.
Currently, we are pulling off this funky stunt to navigate back to the root: https://github.com/owncloud/phoenix/blob/master/apps/files/src/components/FilesTopBar.vue#L6

felixheidecke:
```json
[
    {"index": 0, "text": "",       "to": "/files/list"}, //Suggestion
    {"index": 1, "text": "Funky",  "to": "/files/list/Funky"},
    {"index": 2, "text": "wefwef", "to": "/files/list/Funky/Demo"}
]
```

DeepDiver1975: sounds reasonable - but unrelated to this PR -> move to own PR

PR #394, opened 2019-07-03
Change colour of text-meta on hover to inverted

LukasHirt:
## Screenshots
image

### DIFF
--- a/src/patterns/OcAutocomplete.vue
+++ b/src/patterns/OcAutocomplete.vue
@@ -249,7 +249,8 @@ export default {
     <div class="uk-card uk-card-default uk-card-small uk-card-body uk-margin-top">
       <oc-autocomplete v-model="complexSelection" :items="complexItems" :filter="filterComplexItems" placeholder="type 'er' for example results">
         <template v-slot:item="{item}">
-          <span>{{ item.forename }} {{ item.surname }} <em>(Age: {{ item.age }})</em></span>
+          <span class="uk-text-bold">{{ item.forename }} {{ item.surname }}</span>
+          <div class="uk-text-meta">(Age: {{ item.age }})</div>
         </template>
       </oc-autocomplete>
       <div class="uk-background-muted uk-padding-small uk-margin-small-top">
--- a/src/styles/theme/oc-autocomplete.scss
+++ b/src/styles/theme/oc-autocomplete.scss
@@ -48,5 +48,11 @@
       padding: $global-small-gutter;
       background-color: $global-muted-background;
     }
+
+    &:hover .uk-text-meta,
+    &:focus .uk-text-meta,
+    &-selected .uk-text-meta {
+      color: $global-inverse-color;
+    }
   }
 }


src/styles/theme/oc-autocomplete.scss
```diff
@@ -48,5 +48,11 @@
       padding: $global-small-gutter;
       background-color: $global-muted-background;
     }
+
+    &:hover .uk-text-meta,
```
DeepDiver1975: since we are applying our own logic/styling to the class uk-text-meta: would it make sense to create our own class oc-text-meta?

LukasHirt: Not really sure if it's needed in this case when we apply only one style. In case we would add more then I'd definitely agree. Or am I just missing some other place were we define other styles for this class?

src/patterns/OcAutocomplete.vue
```diff
@@ -249,7 +249,8 @@ export default {
     <div class="uk-card uk-card-default uk-card-small uk-card-body uk-margin-top">
       <oc-autocomplete v-model="complexSelection" :items="complexItems" :filter="filterComplexItems" placeholder="type 'er' for example results">
         <template v-slot:item="{item}">
-          <span>{{ item.forename }} {{ item.surname }} <em>(Age: {{ item.age }})</em></span>
+          <span class="uk-text-bold">{{ item.forename }} {{ item.surname }}</span>
+          <div class="uk-text-meta">(Age: {{ item.age }})</div>
```
felixheidecke: `uk-text-meta` brings it's own color, so just switch to `uk-text-small` and the problem solves itself.

src/styles/theme/oc-autocomplete.scss
felixheidecke: You can drop the whole SCSS if you follow what I mentioned above. Generally, `uk-light` and `uk-dark` (https://getuikit.com/docs/inverse) is the right approach for switching colors.

LukasHirt: `uk-light` doesn't overwrite colour from `uk-text-meta`

felixheidecke:
Bildschirmfoto 2019-07-05 um 13 08 25

Here is what I mean. `uk-light` does change the color of it's elements (must be in the parent wrapper) If the color is too damp, we'd better change the `uk-light` coloring instead of the `uk-text-meta` as UIkit intends.

LukasHirt:
> Bildschirmfoto 2019-07-05 um 13 08 25
> 
> Here is what I mean. `uk-light` does change the color of it's elements (must be in the parent wrapper) If the color is too damp, we'd better change the `uk-light` coloring instead of the `uk-text-meta` as UIkit intends.

Try pls to do this in the autocomplete component. It simply doesn't help there whatever parent wrapper you assign the class.

image

felixheidecke:
I found the problem … `.uk-light` does not exist. I did the same setup in ODS https://github.com/owncloud/owncloud-design-system/pull/394#issuecomment-508724385

Doesn't work. It's not a CSS overriding bug or something similar, uk-light simply doesn't exist 😟

DeepDiver1975:
we take this as a temporary solution

PR #414, opened 2019-07-10
Allow closing of drop on click

LukasHirt:
## Description
Added close on click method and prop to drop component.

## Motivation
We can now use drop for select without the need to create its own component.

### DIFF
--- a/src/elements/OCDrop.vue
+++ b/src/elements/OCDrop.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :uk-drop="$_ocDrop_props">
+  <div :id="dropId" :uk-drop="$_ocDrop_props" @click="$_ocDrop_close">
     <div v-if="$slots.default" class="uk-card uk-card-default uk-card-small uk-card-body">
       <slot />
     </div>
@@ -7,6 +7,8 @@
   </div>
 </template>
 <script>
+import UiKit from "uikit"
+
 /**
  * Position any element in relation to another element.
  */
@@ -15,6 +17,13 @@ export default {
   status: "review",
   release: "1.0.0",
   props: {
+    /**
+     * Id of the drop.
+     */
+    dropId: {
+      type: String,
+      required: false,
+    },
     /**
      * CSS selector of the element to maintain the drop's visibility. see https://getuikit.com/docs/drop#boundary
      **/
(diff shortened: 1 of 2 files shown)


src/elements/OCDrop.vue
```diff
@@ -1,12 +1,15 @@
...
+  <div :id="$_ocDrop_id" :uk-drop="$_ocDrop_props" @click="$_ocDrop_close">
     <div v-if="$slots.default" class="uk-card uk-card-default uk-card-small uk-card-body">
       <slot />
     </div>
     <slot v-else name="special" />
   </div>
 </template>
 <script>
+import UiKit from "uikit"
+import { uniqueId as _uniqueId } from "lodash"
```
DeepDiver1975: is it really necessary to import lodash?

src/elements/OCDrop.vue
```diff
@@ -1,12 +1,15 @@
 <template>
-  <div :uk-drop="$_ocDrop_props">
+  <div :id="$_ocDrop_id" :uk-drop="$_ocDrop_props" @click="$_ocDrop_close">
```
DeepDiver1975: I'd really love to see this more in a vue.js manner to be implemented:
- drop is visible/hidden using v-if or v-show
- clicking the button changes the state upon the drop is shown

We need to get rid of this uikit magic .. one day ...

LukasHirt: Agree.

src/elements/OCDrop.vue
LukasHirt: I haven't found an alternative for the same function outside of downloading some other packages. One way could be using random number but there is the chance of generating two times the same number and then 💥 There is this function exported outside of lodash and distributed as own package but I'm not sure if we want to get it this way.

DeepDiver1975: Bloody uikit

DeepDiver1975: What about setting the ID via a property and leave it to the Developer who uses this component?
Default could be just oc-drop-default-id or Something ....

LukasHirt: Sounds reasonable... Will do :wink:

LukasHirt: Done

Issue #415, opened 2019-07-10
Support event modifiers on our components


DeepDiver1975:
```js
          <oc-checkbox class="uk-margin-small-left" @change.stop="$emit('toggle', item)" :value="selectedFiles.indexOf(item) >= 0" />
```
The stop modifier is not respected.

here are all modifiers: https://vuejs.org/v2/guide/events.html#Event-Modifiers

PR #416, opened 2019-07-10
oc-checkbox: stop event propagation of the click event

DeepDiver1975:
related to #415 - not really fixing it - just to make it work as needed in phoenix as of now


DeepDiver1975:
THX

PR #453, opened 2019-08-02
Added xsmall margin

LukasHirt:
## Motivation
UIkit provides the small margin as the smallest class. In some cases this is not enough and xsmall margin is needed.

### DIFF
--- a/src/styles/theme/margin.scss
+++ b/src/styles/theme/margin.scss
@@ -1,0 +1,21 @@
+@mixin hook-margin-misc() {
+  .uk-margin-xsmall {
+    margin-bottom: $xsmall-margin;
+  }
+  * + .uk-margin-xsmall {
+    margin-top: $xsmall-margin !important;
+  }
+
+  .uk-margin-xsmall-top {
+    margin-top: $xsmall-margin !important;
+  }
+  .uk-margin-xsmall-bottom {
+    margin-bottom: $xsmall-margin !important;
+  }
+  .uk-margin-xsmall-left {
+    margin-left: $xsmall-margin !important;
+  }
+  .uk-margin-xsmall-right {
+    margin-right: $xsmall-margin !important;
+  }
+}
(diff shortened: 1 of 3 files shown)


src/styles/theme/margin.scss
```diff
@@ -0,0 +1,16 @@
+$margin-sizes: (
+  "xsmall": $xsmall-margin,
+);
+$margin-directions: "top", "right", "bottom", "left";
+
+@each $name, $size in $margin-sizes {
+  .oc-margin-#{$name} {
```
felixheidecke: It's rather confusing mixing `uk-margin` with an additional `oc-margin` … I'd rather have it be an uk-margin-… instead.

src/styles/theme/margin.scss
```diff
@@ -0,0 +1,16 @@
...
+    margin: $size;
```
felixheidecke: Add `!important` as well, as seen in the UK Sourcefiles

src/styles/theme/margin.scss
```diff
@@ -0,0 +1,16 @@
...
+
+    @each $direction in $margin-directions {
+      &-#{$direction} {
+        margin-#{$direction}: $size;
```
felixheidecke: `!important` as well

src/styles/theme/margin.scss
felixheidecke: This must be `margin-bottom`

LukasHirt: Why does it need to be `margin-bottom`?

felixheidecke: `uk-margin` works by only adding a margin to the bottom. Only if an element preseeds the margin, then and margin-top is added as in `* + .uk-margin…`

Otherwise, margins would start to stack on certain occasions such as these:

```html
<div class="uk-margin"></div>
<div>
   <div class="uk-margin"></div>
</div>
```

PR #467, opened 2019-08-21
feature/datepicker

felixheidecke:
Select a `date`, `datetime` or simply the `time`.

datepicker

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,6 +151,7 @@
     "jest-haste-map": "^24.8.1",
     "jest-resolve": "^24.8.0",
     "jest-serializer-vue": "^2.0.2",
+    "luxon": "^1.17.2",
     "node-notifier": "^5.4.0",
     "node-sass": "^4.12.0",
     "npm-run-all": "^4.1.5",
@@ -170,6 +171,7 @@
     "theo": "^8.1.3",
     "uikit": "^3.1.6",
     "url-loader": "^2.0.1",
+    "vue-datetime": "^1.0.0-beta.10",
     "vue-jest": "^3.0.4",
     "vue-loader": "^15.7.0",
     "vue-style-loader": "^4.1.2",
@@ -180,7 +182,8 @@
     "webpack-cli": "^3.3.5",
     "webpack-dev-server": "^3.7.2",
     "webpack-merge": "^4.2.1",
-    "webpack-merge-and-include-globally": "^2.1.16"
+    "webpack-merge-and-include-globally": "^2.1.16",
+    "weekstart": "^1.0.0"
   },
   "engines": {
     "node": ">= 8.6.0",
--- a/src/elements/OcDatepicker.vue
+++ b/src/elements/OcDatepicker.vue
@@ -1,0 +1,162 @@
+<template>
+  <datetime
+    v-model="update"
+    class="oc-datepicker"
+    :title="title"
+    :type="type"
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -151,6 +151,7 @@
     "jest-haste-map": "^24.8.1",
     "jest-resolve": "^24.8.0",
     "jest-serializer-vue": "^2.0.2",
+    "luxon": "^1.17.2",
```
DeepDiver1975: What for is luxon needed?

src/elements/OcDatepicker.vue
```diff
@@ -0,0 +1,163 @@
...
+     */
+    date: {
+      type: String,
+      default: moment().format(),
+    },
+    /**
+     * Popup title.
+     */
+    title: {
+      type: [Boolean, String],
```
DeepDiver1975: Why boolean? Use null or empty String to Not Set the Titel.

src/elements/OcDatepicker.vue
```diff
@@ -0,0 +1,163 @@
...
+      default: false,
```
DeepDiver1975: ''

src/elements/OcDatepicker.vue
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Prop definitions should be as detailed as possible, specifying at least
+   * type(s). See examples below:
+   */
+  props: {
+    /**
+     * The html element name used for the container of Example component.
+     */
+    date: {
+      type: String,
```
DeepDiver1975: Isnt there a Datetime type?

src/elements/OcDatepicker.vue
```diff
@@ -0,0 +1,163 @@
...
+      validator: value => {
+        return typeof value.ok === "string" || typeof value.cancel === "string"
+      },
+    },
+    /**
+     * Minimum datetime.
+     * `ISO 8601 String`
+     */
+    minDatetime: {
+      type: [Object, String],
```
DeepDiver1975: Object?

src/elements/OcDatepicker.vue
```diff
@@ -0,0 +1,163 @@
...
+      default: null,
+    },
+    /**
+     * Maximum datetime.
+     * `ISO 8601 String`
+     */
+    maxDatetime: {
+      type: [Object, String],
```
DeepDiver1975: Object?

package.json
felixheidecke: The datepicker uses luxon (instead of moment) for date-"stuff". For some reason, it's not part of the dependencies, so it has to be added manually.

See https://www.npmjs.com/package/vue-datetime#bundler-webpack-rollup

src/elements/OcDatepicker.vue
felixheidecke: Yes, but vue-datetime expects a string here:
https://www.npmjs.com/package/vue-datetime#parameters

src/elements/OcDatepicker.vue
felixheidecke: `null` is an object type. vue-datetime expects a string or null.

src/elements/OcDatepicker.vue
felixheidecke: See above.

PR #481, opened 2019-09-12
Borderless buttons + using a star inside of it

PVince81:
Allow the "link" variation on oc-button for borderless buttons. Note
that buttons are needed for accessibility tools.
Fix oc-star to not render two spans as it would trigger the CSS rule
".oc-button .oc-icon + span" for the second span which would add an
unwanted left margin.

This is for https://github.com/owncloud/phoenix/issues/1645.

### DIFF
--- a/src/elements/OCStar.vue
+++ b/src/elements/OCStar.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="oc-star">
-    <oc-icon v-show="shining" class="oc-star-shining" name="star" />
-    <oc-icon v-show="!shining" class="oc-star-dimm" name="star_border" />
+    <oc-icon v-if="shining" class="oc-star-shining" name="star" />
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
Icons are Design elements which have to be defined in the Design system.
This is how a Design system works.

.... Uikit .... 🤪🤯

PVince81:
@marcus-herrmann do you have an idea if the above properly fits in the design system in this form or whether it should be done differently ?

Please note the original ticket https://github.com/owncloud/phoenix/issues/1645 where it was about having the star clickable for accessibility reasons

PVince81:
@marcus-herrmann your proposal sounds good

PVince81:
seems one can use `variation=raw` in `oc-button`. I missed that, closing.

PR #485, opened 2019-09-17
Auto-select input field contents on focus

PVince81:


PVince81:
This behavior will be useful for example when renaming files/folders so the field contents is preselected so the user can either expand it or replace it.

Issue #443, opened 2019-07-29
Hide icons from assistive technology


LukasHirt:
## Description
Icon labels are often times unnecessary if a button is already sufficiently labeled. Hide the icons from assistive technology (not visually!) with `aria-hidden="true"` Needs to be optional anyway since sometimes only an icon is shown.

LukasHirt:
We no longer set icons via buttons props so this is now solely the responsibility of the devs. It is documented so closing.

PR #496, opened 2019-09-30
 Hide icons from assistive technology, #443

marcus-herrmann:
Does OC have contribution guidelines, regarding branch naming, referencing issues, commit message phrasing, pull request descriptions, min/max amount of reviewers, etc.?

### DIFF
--- a/src/patterns/OcNavbar.vue
+++ b/src/patterns/OcNavbar.vue
@@ -43,8 +43,8 @@ export default {
 ```jsx
   <oc-navbar tag="header" class="oc-topbar">
     <oc-navbar-item position="left">
-      <button class="oc-topbar-menu-burger" aria-label="Files">
-        <oc-icon name="menu" class="oc-topbar-menu-burger-icon" /><span
+      <button class="oc-topbar-menu-burger">
+        <oc-icon name="menu" class="oc-topbar-menu-burger-icon" aria-hidden="true" /><span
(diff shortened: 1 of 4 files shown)


src/patterns/OcNavbar.vue
```diff
@@ -43,8 +43,8 @@ export default {
 ```jsx
   <oc-navbar tag="header" class="oc-topbar">
     <oc-navbar-item position="left">
-      <button class="oc-topbar-menu-burger" aria-label="Files">
-        <oc-icon name="menu" class="oc-topbar-menu-burger-icon" /><span
+      <button class="oc-topbar-menu-burger">
```
DeepDiver1975: from my pov this is wrongly design from a component perspective.
oc-button shall be used. oc button already has the option to have an icon inside and within such an oc-button component the aria labels shall be set properly.

correct my if I'm wrong ;-)

marcus-herrmann: aria-label is not necessary if the button has got an accessible label from another source. In this case it's `<button>this very text here</button>`. Therefore I removed the aria-label attribute. `<button aria-label="foo">foo</button>` would just be redundant.

I did not establish the direct use of `<button />`, if you're referring to this. Assumed it had to be this way here.

LukasHirt:
We no longer set icons via buttons props so this is now solely the responsibility of the devs. It is documented so closing.

PR #497, opened 2019-10-04
Make close button aria-label configurable, for phoenix #1650

marcus-herrmann:
Fixes in part https://github.com/owncloud/phoenix/issues/1650

### DIFF
--- a/src/patterns/OcAppSideBar.vue
+++ b/src/patterns/OcAppSideBar.vue
@@ -7,6 +7,7 @@
             <oc-button
               icon="close"
               class="uk-float-right"
+              :aria-label="closeButtonLabel"
               @click="$emit('close', $event)"
             ></oc-button>
           </slot>
@@ -40,6 +41,10 @@ export default {
     disableAction: {
       default: false,
       type: Boolean,
+    },
+    closeButtonLabel: {
+      default: "Close",
+      type: String,
     },
   },
 }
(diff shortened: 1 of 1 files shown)


src/patterns/OcAppSideBar.vue
```diff
@@ -41,6 +42,10 @@ export default {
       default: false,
       type: Boolean,
     },
+    closeButtonLabel: {
```
LukasHirt: Add pls docs what does this do and why is it needed. e. g. like in https://github.com/owncloud/owncloud-design-system/blob/0db4c32de7fa12a03de7b6aa0d20f09f11aa9ebd/src/elements/OCAvatar.vue#L27

src/patterns/OcAppSideBar.vue
```diff
@@ -41,6 +42,15 @@ export default {
       default: false,
       type: Boolean,
     },
+    /**
+     * Provide the accessible name for the close button, since it has no visible label
+     * but a svg, to which to provide an accessible name would be more complicated.
+     * Value of closeButtonLabel will eventually be the value of the button's aria-label
+     **/
+    closeButtonLabel: {
+      default: "Close",
```
LukasHirt: This string will need to be translated inside Phoenix.

marcus-herrmann: This is considered, see: https://github.com/owncloud/phoenix/blob/bugfix/appmenu-close-btn-label/src/components/Menu.vue#L24

PR #501, opened 2019-10-07
Add accessibility documentation and attributes for OcSpinner

marcus-herrmann:
This PR/commit adds:
* role and tabindex attributes to the spinner component
* makes the ariaLabel prop required
* The reasoning behind it, in form of documentation as an example of the accessible use of this component

### DIFF
--- a/src/elements/OcSpinner.vue
+++ b/src/elements/OcSpinner.vue
@@ -1,10 +1,20 @@
 <template>
-  <div :uk-spinner="'ratio:' + ratio" :aria-label="ariaLabel"></div>
+  <div :uk-spinner="'ratio:' + ratio" :aria-label="ariaLabel" tabindex="-1" role="img"></div>
 </template>
 
 <script>
 /**
  * Remote actions can take an undefined portion of time. The spinner gives feedback to the users about an actions being processed.
+ *
+ * ## Accessibility
+ * ### Making spinners accessible for screen readers
+ *
+ * 1. Making them (only!) programmatically focusable with `tabindex="-1"`. **This is already included in the component.**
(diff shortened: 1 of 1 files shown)


michaelstingl:
@marcus-herrmann could you post a screenshot? Not all stakeholders want to build it to have a look?

marcus-herrmann:
Bildschirmfoto 2019-10-07 um 16 32 33

src/elements/OcSpinner.vue
```diff
@@ -1,21 +1,32 @@
 <template>
-  <div :uk-spinner="'ratio:' + ratio" :aria-label="ariaLabel"></div>
+  <div :uk-spinner="'ratio:' + ratio" :aria-label="ariaLabel" tabindex="-1" role="img"></div>
 </template>
 
 <script>
 /**
  * Remote actions can take an undefined portion of time. The spinner gives feedback to the users about an actions being processed.
+ *
+ * # Accessibility
```
LukasHirt: I'd prefer if we wouldn't start with h1. Since we will get later on some more documentation from other categories I think it would be better to have all of them with h2. Also shouldn't be h1 the name of the component in the header? Looking now that it is h2 (But this is a question for different issue/PR. That of course if it really is an issue.)...

```suggestion
 * ## Accessibility
```

src/elements/OcSpinner.vue
```diff
@@ -1,21 +1,32 @@
...
+ * ## Making spinners accessible for screen readers
```
LukasHirt: See above ^

```suggestion
 * ### Making spinners accessible for screen readers
```

src/elements/OcSpinner.vue
marcus-herrmann: Fixed

marcus-herrmann:
Fixed

PR #512, opened 2019-10-14
Scaffold accessibility documentation for OcSearchBar pattern

marcus-herrmann:
# What's in this PR
* New: A (conditional) landmark role of search, including a prop to disable it
* Changed: The existence of the button is not optional anymore, only whether it is visually perceivable is. Therefore I split the existing `button` prop into `buttonHidden` and `buttonLabel` (Hopefully this will not cause a chain reaction in phoenix)
* Documentation/reasoning of both

Please ping me if this change was too radical since it changes existing ways of using components/APIs

### DIFF
--- a/src/patterns/OcSearchBar.vue
+++ b/src/patterns/OcSearchBar.vue
@@ -1,5 +1,10 @@
 <template>
-  <div uk-grid class="oc-search" :class="{ 'oc-search-small': small }">
+  <div
+    :role="isFilter ? undefined : 'search'"
+    uk-grid
+    class="oc-search"
+    :class="{ 'oc-search-small': small }"
+  >
     <div class="uk-width-expand">
       <div class="uk-inline uk-width-1-1">
         <span v-if="icon" class="uk-form-icon">
@@ -22,14 +27,14 @@
         />
       </div>
     </div>
-    <div v-if="button" class="uk-width-auto">
+    <div :class="[{ 'oc-visually-hidden': buttonHidden }, 'uk-width-auto']">
       <oc-button
         class="oc-search-button"
         :class="{ 'uk-button-small': small }"
         variation="primary"
         :disabled="loading || searchQuery.length < 1"
         @click="onSearch"
-        >{{ button }}</oc-button
+        >{{ buttonLabel }}</oc-button
       >
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


src/patterns/OcSearchBar.vue
```diff
@@ -22,21 +27,30 @@
...
   </div>
 </template>
 <script>
 /**
  * The search bar is an input element used for searching server side resources or to filter local results.
+ *
+ * ## Accessibility
+ *
+ * ### Landmark role=search
+ * Given there is only one instance of `<oc-search-bar>` per page/route, this component should communicate its purpose, being the main site search, to screen readers ([explainer of landmark roles](https://www.washington.edu/accessibility/web/landmarks/)). If the component serves as a filter form, it is advised to disable the landmark role via `isFilter="true"`.
```
LukasHirt: What about the case when there would be two search bars but none of them as a filter? Let's say we have a global search and then some search bar that would be only app-related? Should then both have the "search" role?

marcus-herrmann: Technically speaking: if there are two landmark roles of the same kind you are advised to label them in order to make them differentiable.

Speaking in terms of best practices regarding landmarks: `role=search` should only mark **the** main search form for an interface (global in this case). It's kind of hard to theorize it even further, but what's the global search should be consistent across routes/states/contexts ([WCAG 3.2.4 consistent identification](https://www.w3.org/TR/WCAG21/#consistent-identification)).

LukasHirt: Thanks for explanation!

marcus-herrmann:
@PVince81 I'm uncertain since I changed the props (split `button` into `buttonLabel`and `buttonHidden`).

PVince81:
@marcus-herrmann are you able to test this locally and assess if there is breakage for the known buttons ?

marcus-herrmann:
Currently, I try to. Locally, I referenced the commit that changes things in the package.json:

```
    "owncloud-design-system": "https://github.com/owncloud/owncloud-design-system.git#18bbe426752cc776188b9ea31bcdfe052ae1b85a",
```

Then removed the node_modules folder and ran "yarn install". After sucessfull installation a "yarn watch-all" returns:

```
ERROR in ./src/phoenix.js
Module not found: Error: Can't resolve 'owncloud-design-system' in '/sites/owncloud/phoenix/src'
 @ ./src/phoenix.js 49:0-50 56:8-20
 @ multi whatwg-fetch ./src/phoenix.js

ERROR in ./src/phoenix.js
Module not found: Error: Can't resolve 'owncloud-design-system/dist/system/system.css' in '/sites/owncloud/phoenix/src'
 @ ./src/phoenix.js 50:0-55
 @ multi whatwg-fetch ./src/phoenix.js
```

I attribute this to something other than the contents of the commit, but am not sure what exactly.
In addition to that I am not yet able to run automated tests.

PVince81:
@marcus-herrmann did you manage to retest this now that you are more familiar how to bring the update design system into Phoenix ?

I suggest first rebasing this PR to make sure you got the latest updates

marcus-herrmann:
@PVince81 sucessfully rebased, successfully tested regarding the search region - but noticed there was an implementation detail in Phoenix missing. Related PR for that: https://github.com/owncloud/phoenix/pull/2682/files

PR #513, opened 2019-10-14
Scaffold accessibility documentation for OcNotification/Message components

marcus-herrmann:
# What's in this PR
* New: added `aria-live` and `role` attributes to the part of the notification container that has text inside it
* New: The amount of interrupting the screen reader (assertive, polite) is indirectly controllable with `<oc-notification-message>`'s `status`: Only "danger" creates an assertive screen reader notification that interrupts the screen reader output
* Changed: The close button is now an actual button (instead of link)
* Changed: The close button has an accessible name of "Close" 
* Added: Documentation

### DIFF
--- a/src/elements/OCNotificationMessage.vue
+++ b/src/elements/OCNotificationMessage.vue
@@ -1,7 +1,11 @@
 <template>
   <div class="oc-alert" :class="$_ocNotificationMessage_classes">
     <oc-icon :variation="status" size="large" name="info" class="uk-margin-small-right"></oc-icon>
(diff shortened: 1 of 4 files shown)


PVince81:
@LukasHirt @marcus-herrmann is this mergable now ?

marcus-herrmann:
Should be fine since I did not change the components API

PVince81:
@marcus-herrmann can you resolve the conflict and retest ? then I think we can merge this

marcus-herrmann:
@PVince81 Resolved and retested in ODS

PR #519, opened 2019-10-20
Adjust appbar to a variable height

Julian1998:
Adjust a min-height instead of a fixed height to all app bars. This is needed on narrow screens

Bildschirmfoto von »2019-10-20 14-59-27«
Bildschirmfoto von »2019-10-20 14-59-35«


Julian1998:
needed for https://github.com/owncloud/phoenix/pull/2297

LukasHirt:
This will then break layout in IE 11. There is needed a fixed height.

PR #522, opened 2019-10-21
Add OcActionDrop and OcDisclosure dropdown components...

marcus-herrmann:
...as two accessible alternatives for the OcDrop component.

# What is in this PR?

This PR adds two new components, `OcDisclosureDrop` and `OcActionDrop` which are meant to be substitutes for the `OcDrop` components which has accessibility flaws.

# Why are two seperate components even necessary?

Dropdown and menu are not properly defined words. Sometimes the author/dev just want to make a container (in)visible on click (expample: Phoenix's filter button in the header), sometimes they would supply a list of actions that are comparable to actual operating system menus. Since a web-app aims to deliver an OS-like experience there are occasions where assistive technology should be informed that only actions are waiting within the menu, should be adviced change into and application mode and it should be preannounced that focus is about to be changed programmatically (expample: Phoenix's "New" button in the header). While both components aren't distinguishable on a visual level, both have their specific purposes.

# Why has the disclosure one more aria attributes (like haspopup) and key event listeners?

As documented in the components themselves they follow different patterns (disclosure widget and action menu). Both have a common DOM structure (the container directly follows the button node in the HTML), but `OcActionDrop` communicates to screen readers that they should switch into application mode once the action menu is open. In an action menu, certain keyboard usage patterns are expected (see the component's documentation and [the requirements in theunderlying authoring practice]).

# When should I use the one or the other?

`OcActionDrop` should be used very rarely and also if the dropdown contents are exclusively menuitems/actionitems (please note that as content of an action menu dropdown only nodes with `role="menuitem"` are allowed, and everything else in there, even focusable items, will be hidden from screen readers). But to make menuitems focusable for visual keyboard users, please add `tabindex="0"` to them (see example code and documentation).

# Non-owncloud Examples
### Slack
Slacks kebap menu

### Gmail
Gmail's settings menu

### DIFF
--- a/src/elements/OcActionDrop.vue
+++ b/src/elements/OcActionDrop.vue
@@ -1,0 +1,180 @@
+<template>
+  <div class="oc-dropdown-wrapper" ref="menu" @keyup.esc="closeHandler(true)">
+    <button
+      :aria-expanded="isOpen.toString()"
+      :class="$_ocButton_buttonClass"
+      ref="button"
+      aria-haspopup="true"
+      @click="toggleDropdown"
+      @keydown.enter.prevent="openMenuAndFocus(0)"
+      @keydown.space.prevent="openMenuAndFocus(0)"
+      @keydown.down.prevent="openMenuAndFocus(0)"
+      @keydown.up.prevent="openMenuAndFocus(-1)"
+    >
+      <slot name="button"></slot>
+    </button>
+    <div
+      class="uk-card-default oc-dropdown-menu"
+      role="menu"
+      ref="dropdown"
+      :hidden="!isOpen"
+      @click="closeHandler"
+      @keydown.down.prevent="focusNext"
+      @keydown.up.prevent="focusPrev"
+    >
+      <slot name="actions"></slot>
+    </div>
+  </div>
+</template>
+
+<script>
+/**
+ * Action dropdowns are meant to be used for dropdown menus that are emulating desktop app menus. That means: only consist of items that change state in the app, or open modal dialogs, for example. For using links, inputs, etc. in dropdowns use the disclosure widget pattern (OcDisclosureDrop).
+ *
+ * ## Accessibility
+ *
+ * ### DOM structure
+ * For keyboard and screen reader focus reasons the dropdown's container must follow the menu button directly in the DOM.
+ *
+ * ### Aria roles
+ * OcActionDrop uses the [ARIA Authoring Practices' action menu pattern](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions.html). The dropdown itself has `role="menu"`, its children need to have `role="menuitem"`. The triggering button has the attribute `aria-haspopup="true", which indicates to assitive technology that it opens an action menu. When a screen reader discovers an action menu (and supports the concept of different input modes, like NVDA and JAWS) it switches to application mode.
+ *
+ * ### Allowed content for the dropdown
+ * Quote from [https://accessibility.dev/challenges-mega-menus-standard-menus-make-accessible/](https://accessibility.dev/challenges-mega-menus-standard-menus-make-accessible/):
+ *
+ * > The use of ARIA Menu roles prohibits the inclusion of any non-menu active element roles within the same structure.
+ *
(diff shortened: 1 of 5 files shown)


src/elements/OcDisclosureDrop.vue
```diff
@@ -0,0 +1,149 @@
...
+    window.addEventListener("keyup", e => {
+      if (e.key === "Escape") {
+        this.closeHandler()
+      }
+    })
+  },
+}
+</script>
+
+<style scoped>
```
LukasHirt: Move styles pls into its own sass file.

src/elements/OcActionDrop.vue
```diff
@@ -0,0 +1,208 @@
...
+      }
+    })
+  },
+  updated: function() {
+    this.focusables = this.$refs.dropdown.querySelectorAll(this.focusableElements)
+  },
+}
+</script>
+
+<style scoped>
```
LukasHirt: Move styles pls into its own sass file.

src/elements/OcActionDrop.vue
```diff
@@ -0,0 +1,208 @@
+<template>
+  <div class="wrapper" ref="menu" @keyup.esc="closeHandler(true)">
```
LukasHirt: Let's rename this to some 'oc-' class? To reflect what we're doing with other components.

src/elements/OcDisclosureDrop.vue
```diff
@@ -0,0 +1,149 @@
+<template>
+  <div class="wrapper" ref="menu" @keyup.esc="closeHandler(true)">
```
LukasHirt: Let's rename this to some 'oc-' class? To reflect what we're doing with other components.

marcus-herrmann:
@LukasHirt Done. Also, added refinements.

src/elements/OcActionDrop.vue
```diff
@@ -0,0 +1,180 @@
...
+      :class="$_ocButton_buttonClass"
+      ref="button"
+      aria-haspopup="true"
+      @click="toggleDropdown"
+      @keydown.enter.prevent="openMenuAndFocus(0)"
+      @keydown.space.prevent="openMenuAndFocus(0)"
+      @keydown.down.prevent="openMenuAndFocus(0)"
+      @keydown.up.prevent="openMenuAndFocus(-1)"
+    >
+      <slot name="button"></slot>
```
LukasHirt: Is there a reason why it's slot and not a prop?

marcus-herrmann: Yes, so could put texts, (labelled) icons, images (with alt-text describing the function of the button), media, or what-not in there. Or style text. Slots are more flexible, especially since these components aim to be generic.

LukasHirt: Text and icons can be both handled via props. Haven't thought about an image though - think it would even break the layout of button now. But yeah, you're right - slot probably suits this better.

src/elements/OcDisclosureDrop.vue
```diff
@@ -0,0 +1,117 @@
...
+      this.isOpen = false
+    },
+  },
+  created() {
+    // Add a general event listener to catch outside clicks
+    document.addEventListener("click", this.documentClick)
+  },
+  destroyed: function() {
+    // Remove the general event listener to catch outside clicks
+    document.removeEventListener("keyup", this.documentClick)
```
LukasHirt: Keyup?

src/elements/OcDisclosureDrop.vue
```diff
@@ -0,0 +1,117 @@
...
+  },
+  mounted() {
+    // Add ESC key event listener
+    window.addEventListener("keyup", e => {
```
LukasHirt: I feel a little bit scared seeing this event which is not then removed.

src/elements/OcDisclosureDrop.vue
marcus-herrmann: Typo. Thanks for noticing. Resolved

src/elements/OcDisclosureDrop.vue
marcus-herrmann: Updated: will get removed `beforeDestroy`

PVince81:
need conflict resolution

marcus-herrmann:
@PVince81 resolved

PR #524, opened 2019-10-28
Various accessibility improvements regarding labels

marcus-herrmann:
# What's in this PR

* Make labels required. Unfortunately there is no way to check for the slot's content in prop validation. Vue Docs: "Note that props are validated before a component instance is created, so instance properties (e.g. data, computed, etc) will not be available inside default or validator functions."
* Distinguish between placeholder and label
* Improve example code with labels
* Prevent explicit and aria-label at the same time

### DIFF
--- a/src/elements/OcCheckbox.vue
+++ b/src/elements/OcCheckbox.vue
@@ -1,11 +1,14 @@
 <template>
-  <input
-    class="oc-checkbox"
-    type="checkbox"
-    v-model="checkboxState"
-    @change="$_ocCheckbox_change($event)"
-    @click.stop
-  />
+  <label>
+    <input
+      class="oc-checkbox"
+      type="checkbox"
+      v-model="checkboxState"
+      @change="$_ocCheckbox_change($event)"
+      @click.stop
+    />
+    <span v-text="label" :class="{ 'oc-visually-hidden': labelVisuallyHidden }" />
+  </label>
 </template>
 <script>
 /**
@@ -20,6 +23,22 @@ export default {
      * Data-model
      **/
     value: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
+    /**
+     * Label of the Checkbox
+     **/
+    label: {
+      type: String,
+      required: true,
+      default: null,
+    },
+    /**
+     * Is the label of the Checkbox visually hidden?
+     **/
+    labelVisuallyHidden: {
       type: Boolean,
       required: false,
       default: false,
@@ -60,10 +79,10 @@ export default {
                 <h3 class="uk-heading-divider">
                     Checkboxes Types
                 </h3>
-                <oc-checkbox />
-                <oc-checkbox v-model="checkState1"/>
+                <oc-checkbox label="Label" />
+                <oc-checkbox v-model="checkState1" label="Label"/>
                 <div @click="parentClick">
-                    <oc-checkbox />
+                    <oc-checkbox label="Label" />
                 </div>
             </section>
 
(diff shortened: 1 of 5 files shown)


src/elements/OcCheckbox.vue
```diff
@@ -1,11 +1,14 @@
...
-  />
+  <label>
+    <input
+      class="oc-checkbox"
+      type="checkbox"
+      v-model="checkboxState"
+      @change="$_ocCheckbox_change($event)"
+      @click.stop
+    />
+    <span :class="labelVisuallyHidden ? 'oc-visually-hidden' : ''">{{ label }}</span>
```
LukasHirt: Change pls to `v-text`

src/elements/OcRadio.vue
```diff
@@ -1,14 +1,7 @@
 <template>
   <label class="radio-container">
     <span class="label">{{ label }}</span>
```
LukasHirt: Change this pls to `v-text`.

src/elements/OcTextarea.vue
```diff
@@ -15,6 +15,9 @@
  * Textareas are used to allow users to provide text input when the expected
  * input is long. Textarea has a range of options. For shorter input,
  * use the `Input` element.
+ *
+ * # Accessibility
```
LukasHirt: Move pls one level.

src/elements/OcTextInput.vue
```diff
@@ -4,7 +4,7 @@
     :type="type"
     :value="value"
     :placeholder="placeholder"
-    :aria-label="placeholder"
+    :aria-label="label"
```
LukasHirt: Let's move here pls also docs about why is this different as in text area component.

marcus-herrmann: In what regard different? both have the value of placeholder mapped to the placeholder attribute; and the value of label to the aria-label attribute.

src/elements/OcTextarea.vue
marcus-herrmann: Done

src/elements/OcRadio.vue
marcus-herrmann: Done

src/elements/OcTextInput.vue
LukasHirt: The docs about the placeholder vs. label. Why the placeholder shouldn't be used as a label. "The attributes `placeholder` and `aria-label` have different functions..." As in text area.

marcus-herrmann: Got it. Added that bit.

LukasHirt: Thanks!

src/elements/OcTextarea.vue
LukasHirt: Thanks

src/elements/OcRadio.vue
LukasHirt: Thanks

src/elements/OcCheckbox.vue
```diff
@@ -1,11 +1,14 @@
...
-  />
+  <label>
+    <input
+      class="oc-checkbox"
+      type="checkbox"
+      v-model="checkboxState"
+      @change="$_ocCheckbox_change($event)"
+      @click.stop
+    />
+    <span v-text="label" :class="labelVisuallyHidden ? 'oc-visually-hidden' : ''" />
```
LukasHirt: This way you can drop the ternary operator.

```suggestion
    <span v-text="label" :class="{ 'oc-visually-hidden': labelVisuallyHidden }'" />
```

marcus-herrmann: Cheers. Edited.

PVince81:
missing bits: https://github.com/owncloud/owncloud-design-system/pull/587

PR #525, opened 2019-10-28
Refactor and document OcStar(Toggle)

marcus-herrmann:
# What's in this PR
* Simplification of OcStar
* New scss helper function for "raw" buttons
* New: OcStarToggle component as a button `role="switch"` conveying its state with `aria-checked="true|false"

# OcStarToggle as a replacement for OcSwitch
In ODS OcSwitch is an image that reacts to a toggle button, but is not the toggle button itself. OcStarToggle matches Phoenix's current implementation of the star being the interactive control (alas, not a button. But this is what this OR is for).

# Relation to issues
Discussion started in this PR: https://github.com/owncloud/owncloud-design-system/pull/481#issuecomment-544442414

# Background on toggle/switch buttons:
[Article by Heydon Pickering, longread](https://inclusive-components.design/toggle-button/#thisdoesntquitefeelright)

### DIFF
--- a/src/elements/OcStarToggle.vue
+++ b/src/elements/OcStarToggle.vue
@@ -1,0 +1,65 @@
+<template>
+  <oc-button
+    variation="raw"
+    role="switch"
+    :aria-checked="checked.toString()"
+    :aria-label="label"
+    @click="$_ocStarToggle_togglePressed"
+  >
+    <oc-star :shining="checked" />
+  </oc-button>
+</template>
+
+<script>
+import OcButton from "./OcButton"
+
+/**
+ *
+ * ## Accessibility
+ *
+ * ### Communicating the buttons toggle role to screen readers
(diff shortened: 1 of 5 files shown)


src/elements/OcStarToggle.vue
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * */
+
+export default {
+  name: "oc-star-toggle",
+  components: { OcButton },
+  status: "review",
+  release: "1.0.0",
+  methods: {
+    togglePressed: function() {
```
LukasHirt: Add `$_ocStarToggle` prefix pls to keep the naming convention in ODS. (see e. g. https://github.com/owncloud/owncloud-design-system/blob/5cf4b0bdc69a9f32bddc2d90a082e08f9e961938/src/patterns/OcAutocomplete.vue#L214)

PR #532, opened 2019-11-07
Feature/new menu topbar idea 1

DeepDiver1975:
Screenshot from 2019-11-07 13-11-49

### DIFF
--- a/src/elements/OCAvatar.vue
+++ b/src/elements/OCAvatar.vue
@@ -1,7 +1,7 @@
 <template>
   <oc-img
     class="oc-avatar"
-    width="50"
+    :width="width"
     :src="src"
     :alt="accessibilityLabel"
     :aria-hidden="accessibilityLabel === ''"
@@ -33,6 +33,12 @@ export default {
       type: String,
       required: false,
       default: "",
+    },
+
+    width: {
+      type: Number,
+      required: false,
+      default: 50,
     },
   },
 }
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
Also with app menu like google - as an idea to discuss
Screenshot from 2019-11-07 13-54-33

PVince81:
@wuenschedesign can you review the proposal ? (you might want to ignore the visual flaws as this was done quickly)

I think this is the way to go, we need to separate the files app entries from the actual application entries as it will fit better with the way we want to do Phoenix extensions in the future.

Basically the header and the application menu will be reused also in OC 10 to provide a seamless experience.

PVince81:
I've started splitting the menus already in https://github.com/owncloud/phoenix/pull/2620.
In that one I've kept the left sidebar to slide in as before, not sure if this should change as well in the new design to have it always visible like in OC 10 ?

PVince81:
PR that implements this in Phoenix: https://github.com/owncloud/phoenix/pull/2669

Next todos for this PR:
- [ ] Fix the menu button styles
- [ ] Proper icon for app switcher, also important to be expose it to Phoenix to use it there

DeepDiver1975:
So I guess we no longer need this?
Or shall we create the components (user menu, app switcher, nav bar etc ....) within ods?

PVince81:
We need at least this PR as it contains the "example page" with some markup and styles, which I used as inspiration for the actual implementation.

Not sure if we need smaller design system components, it feels a bit like overkill considering that these components will only be used in one place (in the top-bar vue / web component)

PVince81:
- [ ] revert file list changes and keep only top bar changes

PVince81:
- [ ] need oc-icon with URL support for the app switcher: https://github.com/owncloud/owncloud-design-system/issues/562

PVince81:
Superseded by https://github.com/owncloud/owncloud-design-system/pull/568 which only picks the top bar changes

Issue #535, opened 2019-11-11
Accessible tabs


PVince81:

pascalwengerter:
Closed as outdated since we're removing the (deprecated) tabs components in the near future

PR #539, opened 2019-11-12
Adds new tab component tabbed including child components, for #535

marcus-herrmann:
Establishes:

* New tab component `oc-tabbed`
* New tab-item component `oc-tabbed-tab`
* New panel-item component `oc-tabbed-panel`

Structure according to the briefing by @PVince81 

Keyboard behaviour, ARIA roles, properties and states according to the [Authoring Practice](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.htmll) 

Intention for establishing new components: Trying to avoid to break phoenix by using the `oc-tab-item` namespace since the new component is much more complex than the existing one.

### DIFF
--- a/src/elements/OcTabbed.vue
+++ b/src/elements/OcTabbed.vue
@@ -1,0 +1,138 @@
+<template>
+  <div>
+    <div
+      class="oc-tabbed"
+      role="tablist"
+      ref="tablist"
+      @click="clickHandler"
+      @keydown="keydownHandler"
+    >
+      <slot name="tabs"></slot>
+    </div>
+    <div ref="panellist">
+      <slot name="panels"></slot>
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "oc-tabbed",
+  status: "prototype",
+  release: "1.0.0",
+  data() {
+    return {
+      tabs: [],
+      panels: [],
+      activeTab: `tab-${this.initial}`,
+    }
+  },
+  props: {
+    initial: {
+      type: String,
+      required: true,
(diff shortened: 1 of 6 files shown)


PVince81:
@marcus-herrmann can you test the component in Phoenix here: https://github.com/owncloud/phoenix/blob/master/apps/files/src/components/FileDetails.vue#L31

PVince81:
code looks fine, as discussed.

please submit a PR in Phoenix that works with this PR here, this will confirm that the new tabbed component can work "on the field"

src/elements/OcTabbed.vue
```diff
@@ -0,0 +1,138 @@
...
+        this.$emit("activeTab", tab.replace("tab-", ""))
+        this.$emit("click")
+      })
+    },
+    clickHandler(e) {
+      if (!e.target.attributes["aria-controls"]) return
+      this.changeTab(e.target.id)
+    },
+    keydownHandler: function(e) {
+      const activeElem = this.$refs.tablist.querySelector(`#${this.activeTab}`)
```
DeepDiver1975: using query selector feels do un-vuejs-ish ... this is what vuejs has refs for ..... or am I missing anything?

marcus-herrmann: I can't really use a ref here since the name is dynamic, I think

PVince81: indeed, seems not possible. refs are limited to the current component: https://stackoverflow.com/a/42814625

PR #541, opened 2019-11-12
Introduced selected background colour

LukasHirt:
## Description
Implemented selected background colour. The value of this colour is taken from Figma oC library colours.

## Screenshot
image


LukasHirt:
As mentioned in the desc, value is taken from Figma oC library where are already new colour which has been used e. g. by iOS app and should be brought into Phoenix as well. Was clarified with @wuenschedesign that those colours shouldn't change (or at least not in a big scale)

Issue #537, opened 2019-11-11
Document design guidelines


PVince81:
TODO: copy result of UX workshop here

PVince81:
- [ ] accessibility: use tables and not divs to make tables
- [ ] button name must match action (ex: "Delete" instead of "OK")
- [ ] try to avoid modals as far as possible (check other solutions)
- [ ] decide on styling when folder / file or user is mentionned in sentences / prompts, etc (for example: bold or italic, quotes, ...)
   - [ ] => have a design system component with the styles
   - [ ] => could apply to emails as well
- [ ] always use text in icon buttons except for obvious cases (like a cross for closing)
- [ ] remove button or any controls instead of disabling whenever a feature is not accessible
- [ ] tone of voice, when to use "you" and "me" (open)
- [ ] to work on, how to display errors, validation, etc (open)

@marcus-herrmann above is the result of the workshop discussion
if you feel like some topics are too broad and need discussion, please raise a separate ticket for it and link in the checkbox row

marcus-herrmann:
@PVince81 First draft: https://github.com/owncloud/owncloud-design-system/pull/542

marcus-herrmann:
> @PVince81 First draft: #542

@PVince81 is this (#542) what you had in mind?

pascalwengerter:
Closing this as party fixed / outdated / tracked elsewhere (e.g. #1193 )

PR #542, opened 2019-11-12
Add interface guidelines, for #537

marcus-herrmann:
Results of the workshop, destilled in Interface Guidelines documentation page

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -1,0 +1,35 @@
+## Interface Element Guidelines
+
+### Be empathetic
+
+Have in mind that the interface is not built only for able-bodied people but for all humans. Therefore make sure that it can be used, for example, in an non-visual way (e.g. relying on the DOM created by JavaScript) or with non-pointer devices like mouse and touch (e.g. keyboard-only or single-switch devices). Test your work in as many forms of [assistive technology](https://webaccess.berkeley.edu/resources/assistive-technology) as possible.
+
+### Supply meaning with the right choice of elements
+
+Regardless of the visual representation, choose a meaningful and semantic HTML element when building a component. [Find a list of elements here at MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element). If an interface concept does not find a match in HTML, supply meaning, information about state, role, relation and name of an element via [WAI-ARIA](https://www.w3.org/TR/wai-aria/).
+
+#### Example: Tabular Content
+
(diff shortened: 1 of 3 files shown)


marcus-herrmann:
@PVince81 Is that the type of documentation you had in mind?

docs/guidelines.md
```diff
@@ -0,0 +1,35 @@
...
+
+#### Example: Tabular Content
+
+When outputting a large amount of structured data and related actions to said data are present, grid-like and single sets of data are ordered in columns (like in the `OcFileList` component), whereas single files and folders are ordered in rows, choose the `table` element.
+
+#### Example: Crucial controls
+
+A semantic choice of elements is important on the most vital elements of Phoenix (such as controls related to files and folders), especially for users who do not use a pointing device. Make sure that a control is focusable and conveys its state and function. In the case of the file and folder example mentioned above differentiate between changing the state of the web-app (`<button>`, triggers non-modal dialog, editor) and changing the location (`<a href="/">`, leads to a route, showing contents of a sub folder).
```
PVince81: don't mention Phoenix. the design system could be used elsewhere

docs/guidelines.md
```diff
@@ -0,0 +1,35 @@
...
+
+The use of the "modal dialog" design pattern should be avoided. Its character is to force a certain interface element into the users field of view and focus, while simultaneously disabling all other present interface elements. The purpose of a modal dialog is to force a user decision. Alternatives could be: non-modal dialogs, disclosure widgets, inline-editing, undo functionality.
+
+### Emphasize core entities
+
+If core entities (Files, Folders, Users) are mentioned in the interface's copy (also applying to HTML e-mails sent by the system), emphasize its mentions by text format (bold, italic ) or by putting the mention in quotes.
+
+### Choose hiding over disabling interactive elements
+
+Instead of disabling elements that are not accessible to a user (for example due to a lack of role privileges on their account) hide irrelevant controls altogether.
```
PVince81: doesn't necessarily apply for a "Save" button for example, as the button should be disabled until all required fields and validation errors are solved (I'm mentionning this detail because we had a "Save" button that hid itself before)

marcus-herrmann: Modified both.

marcus-herrmann:
@PVince81 Updated this PR with context why a change of a `role` on an element alone is not advisable

Issue #536, opened 2019-11-11
Accessible uploads


PVince81:

marcus-herrmann:
Goal is the follwing strategy to accomodate improved keyboard and screen reader accessibility:

* After the files have been selected via the "File open" prompt supported by the users operating system, focus should be sent to `<oc-progress>`
* `<oc-progress>` gets `tabindex="-1"` to become programmatically focusable
* `<oc-progress>` should supply information of the progress to screen readers via `aria-valuemin`, `aria-valuemax` and `aria-valuenow`. Since the visual progress bar is labelled in percent values, `aria-valuemin` should be 0, `aria-valuemax` 100 and `aria-valuenow` the current dynamic value. While focus is on the `progress` element, its state will be communicated to supporting screen readers
* After the upload is complete:
   * an [`aria-live` region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) announces that the upload has ended successfully
   * Focus sent to the most resonable place, maybe the `<main>` element

PR #556, opened 2019-11-28
Improve accessibility of OgProgress and add live region component, for #536

marcus-herrmann:
Improves screen reader communication of `OcProgress` and adds a generic live region component (boths first use will be https://github.com/owncloud/phoenix/pull/2542).

Double semantics in OcProgress.vue  (`<progress role="progressbar" ... />`) is to remedy a shortcoming in JAWS screenreader.

### DIFF
--- a/src/elements/OcProgress.vue
+++ b/src/elements/OcProgress.vue
@@ -1,5 +1,15 @@
 <template>
-  <progress class="uk-progress" :value="value" :max="max"></progress>
+  <progress
+    :value="value"
+    :max="max"
+    :aria-valuemax="max"
+    :aria-valuenow="value"
+    aria-busy="true"
+    aria-valuemin="0"
+    class="uk-progress oc-progress"
+    tabindex="-1"
+    role="progressbar"
+  ></progress>
 </template>
 <script>
 /**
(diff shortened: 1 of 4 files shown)


src/elements/OcProgress.vue
```diff
@@ -1,5 +1,15 @@
 <template>
-  <progress class="uk-progress" :value="value" :max="max"></progress>
+  <progress
+    :value="value"
+    :max="max"
+    :aria-valuemax="100"
```
PVince81: shouldn't this also be "max" ?

marcus-herrmann: Completely right. Fixed

src/elements/OcProgress.vue
```diff
@@ -1,5 +1,15 @@
 <template>
-  <progress class="uk-progress" :value="value" :max="max"></progress>
+  <progress
+    :value="value"
+    :max="max"
+    :aria-valuemax="max"
+    :aria-valuenow="value"
+    aria-busy="true"
```
PVince81: just noticed this: this assumes that we'll always use `v-if` to avoid rendering the progress bar when not busy.
if one would use `v-show` with a falsy value, would aria-busy still be in effect ?

probably not critical, I'm fine keeping it like this now

marcus-herrmann: To my understanding `aria-busy` is a value that would be in effect when just hidden, but wouldn't do any harm communication-wise as long as the progressbar is not focussed

PR #565, opened 2019-12-11
Fixes #562: the icon can now be loaded from a given URL

### LINKED ISSUE #562: oc-icon shall optionally take a url from where to load the svg to be used
In some cases we need to display an icon as given to us by path - e.g. https://github.com/owncloud/phoenix/pull/2651/files#diff-9e4fc1110432e662806df44bc5b846a5R90

DeepDiver1975:


src/elements/OcIcon.vue
```diff
@@ -24,9 +25,15 @@ export default {
      * The name of the icon to display.
      */
     name: {
-      required: true,
+      type: String,
       default: "settings",
```
LukasHirt: Let's remove the default then? Or put there a condition to use it only if the URL is not specified?

DeepDiver1975: taken care of - THX

PR #587, opened 2019-12-13
Search bar label

PVince81:
Follow up from https://github.com/owncloud/owncloud-design-system/pull/524

Adjusts all other occurrences from oc-input-field


PVince81:
double approval is best approval, thanks y'all :+1:

PR #592, opened 2019-12-18
Fixes #582 - using vue-avatar

### LINKED ISSUE #582: Avatar placeholder
When no URL is given or the URL failed to load, we should fall back to using a letter as an avatar placeholder.

We could have a look at https://github.com/eliep/vue-avatar

DeepDiver1975:
Screenshot from 2019-12-18 09-44-22


PVince81:
some issues: https://github.com/owncloud/owncloud-design-system/issues/596

PR #607, opened 2020-01-09
Share indicators + new icon colors "active" and "passive"

PVince81:
image

This idea brings all share indicators into a separate column.

### Level 1

- which resource is shared: see whether column icons are present
- what kind of share ? (user/link): see icon types in column
- is the share applied from a parent instead of directly ? => see reduced opacity
- optional: incoming/outgoing ? icon color

### Level 2

- who has access ?
   - open right sidebar for resource
   - list of collaborators sums up collaborators of parents (+ via jump)
   - list of links sums up links of parents

## TODOs:

- [x] keep column in responsive mode ?
- [x] check if desktop and mobile clients can align with this design ?
- [x] question: feature request for opening sidebar for current folder ??

### DIFF
--- a/src/elements/OcIcon.vue
+++ b/src/elements/OcIcon.vue
@@ -76,7 +76,7 @@ export default {
       type: String,
       default: "system",
       validator: value => {
-        return value.match(/(file-type|system|inverted|danger|success|warning)/)
+        return value.match(/(file-type|system|active|passive|inverted|danger|success|warning)/)
       },
     },
   },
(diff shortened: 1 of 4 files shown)


PVince81:
@pmaier1 @michaelstingl

PVince81:
Feedback was that we shouldn't use lower opacity, because users would not be able to know that it's "light grey" if there are no other visible icons. So we'll go for colors instead, this will require to adjust "oc-icon" to support color values or to allow styling through classes.

PVince81:
Updated with colors, where only direct outgoing shares are highlighted/active:

image

PVince81:
- [ ] adjust icon colors to better fit the general palette

PVince81:
for more complex cases, see https://github.com/owncloud/phoenix/pull/2877 which makes it interactive

PVince81:
I've shown this several times and people were fine with the current design.

I'd like to move forward and merge this.
We can decide to tweak colors later on when we see it working.

Please review @LukasHirt @kulmann

PVince81:
squashed and rebased

PR #628, opened 2020-01-29
Migrate CI to starlark add Changelog and Github Releases

micbar:
## Description
- Create a new Drone Pipeline for this Repo

## Added
- `yarn lint` step on every event
- Calens Changelog generator
- Calens Test Step on PRs
- Calens Test and commit step for merge commits to master
- Test Docs step on all events
- Test Build step on all events
- Publish Docs step on tag and merge events
- Publish NPM step on tag events
- Create Github release on tag event

## Removed
- yarn version step to adapt the version in package.json This needs to be done manually now.


.drone.star
```diff
@@ -0,0 +1,258 @@
...
+                    'email': {
+                        'from_secret': 'npm_email',
+                    },
+                    'token': {
+                        'from_secret': 'npm_token',
+                    },
+                },
+                'when': {
+                    'ref': {
+                        'exclude': [
```
LukasHirt: Why not use also tags as a ref instead of excluding as done in changelog and release?

LukasHirt:
> Publish Docs step on tag and merge events

I'm thinking now if this shouldn't be done also only on tags not to confuse the developer that would come to the docs, would see version x.x.x with component y and would want to use it but it wouldn't be yet published to npm.

.drone.star
micbar: done

PR #630, opened 2020-01-31
Fix breadcrumb component ellipsis

PVince81:
Added overflow: hidden on the main element of the oc-breadcrumb to make
sure that it can ellipsize its content in responsive mode.

image

Before this fix it would just overflow.

You can also try this in Phoenix by having long folder names and adding "overflow:hidden" on that element and see the difference.

### DIFF
--- a/src/elements/OcBreadcrumb.vue
+++ b/src/elements/OcBreadcrumb.vue
@@ -38,7 +38,7 @@
  *
  *  - text: mandatory element, holds the text which is to be displayed in the breadcrumb
  *  - to: optional element, the vue router link
- *  -
+ *
  */
 export default {
   name: "oc-breadcrumb",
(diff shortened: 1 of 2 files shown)


src/elements/OcBreadcrumb.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="oc-breadcrumb">
+  <div class="oc-breadcrumb uk-overflow-hidden">
```
LukasHirt: Extend the `oc-breadcrumb` class pls with the overflow in its scss file - https://github.com/owncloud/owncloud-design-system/blob/ca7575202ef2009a6d603d660b43db89a3be4b61/src/styles/theme/oc-breadcrumb.scss#L6

PVince81:
@LukasHirt adjusted accordingly, thanks for pointing in the right direction :-)

PR #632, opened 2020-02-06
Add bigger icon sizes

PVince81:
Added "xlarge" with 96px.
Added "xxlarge" with 192px.
Adjusted indent in OcIcon.
Changed icon size presentation to use a tabular layout.

image


PVince81:
In Phoenix empty folder with xlarge:
image

With xxlarge:
image

PR #639, opened 2020-02-13
Added docs about escaping translations

LukasHirt:
image

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -1,3 +1,21 @@
+## Writing translations
+
+For translations we are using combination of [vue-gettext](https://github.com/Polyconseil/vue-gettext) and [easygettext](https://github.com/polyconseil/easygettext).
+
+### Escaping
+
+Translations in combination with [mustache syntax](https://mustache.github.io/) results in double escaping. This leads to an issue where some special characters are not displayed correctly. To prevent this issue in cases where we expect special characters (resource names, user names, etc.), we disable the escaping on the side of translation tools. This way the text gets escaped only once.
+
+For translations where we don't expect special characters, we explicitly enable escaping.
+
+#### Example with disabled escaping
+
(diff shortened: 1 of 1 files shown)


LukasHirt:
@C0rby for some reason I can't add you as a reviewer because it doesn't find you there 🤷‍♂ Pls check these docs as well :pray:

PVince81:
After reading the explanation it doesn't feel clear enough.

Wasn't this about avoiding double escaping ? Is the undesirable result the double escaping ? Might be good to be clear about this, because I usually shiver when I read disabling escaping for user controlled values. So we better make sure we are still on the safe side and say so explicitly (probably the thing about the mustache syntax)

DeepDiver1975:
Is this something for the design system? We have no translations in here and we are also not forcing any project on the tooling.

From my pov this needs to go into the docs for phoenix .... somewhere .....

LukasHirt:
> After reading the explanation it doesn't feel clear enough.

I'll try to go more into details :wink:

LukasHirt:
> Is this something for the design system?

@DeepDiver1975 This was discussed recently and the reasoning behind this is to have the guidelines united and since we are already including accessibility and design ones there (plus some internationalization already as well), seems ok to have there also parts about the actual code.

LukasHirt:
@PVince81 Docs adjusted.

PVince81:
I also have mixed feelings about this but agree that having all in one place is slightly better.
Let's go with that for now.

PVince81:
thinking... adding the docs to Phoenix core will not help getting awareness for when we split Phoenix files and other extensions, as we'd need to duplicate the docs to these repos if they all use easygettext.

and some extensions might not use easygettext at all.

anyway, existing doc is better than no doc at all

PVince81:
raised https://github.com/owncloud/owncloud-design-system/issues/642

PR #649, opened 2020-02-24
Upgraded libs

LukasHirt:


build/utils.js
```diff
@@ -74,18 +74,14 @@ exports.cssLoaders = function(options) {
     },
   }
 
-  const sassOptions = {
```
LukasHirt: Not valid config anymore in the new version.

config/docs.config.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
   editorConfig: {
     theme: "night",
   },
+  simpleEditor: false,
```
LukasHirt: New option coming from new version - needs to be set otherwise the build will fail. https://vue-styleguidist.github.io/Configuration.html#editorconfig

PR #652, opened 2020-02-25
Remove uppercase on buttons

PVince81:
UIKit is adding text-transform: uppercase on all buttons.
This commit removes it in the standard theme.

Not sure if the theme is the right place though...

### DIFF
--- a/src/styles/theme/oc-button.scss
+++ b/src/styles/theme/oc-button.scss
@@ -26,6 +26,8 @@
     margin-left: $global-margin;
   }
 
+  text-transform: none !important;
+
   &:not(.uk-button-small):not(.uk-button-large) {
     min-height: $global-control-height;
   }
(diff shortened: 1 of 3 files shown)


src/styles/theme/oc-button.scss
```diff
@@ -26,6 +26,10 @@
     margin-left: $global-margin;
   }
 
+  & {
```
LukasHirt: Why using `&`? It can be directly in the `.oc-button` styles

PVince81: it didn't work at first, but also I didn't have the `!important`.

trying now...

PVince81: adjusted

PVince81:
@LukasHirt changelog added

Issue #654, opened 2020-02-25
Don't force a oc-button height in raw mode


PVince81:
When using `<oc-button variation="raw" ...>`, no size attribute should be expected and no size should be applied to the resulting markup as it.

PVince81:
PR here https://github.com/owncloud/owncloud-design-system/pull/655

PR #655, opened 2020-02-25
Removed size for oc-button with raw variation

PVince81:
Raw variation of buttons have no border, so they now also don't have a size enforced to avoid needless white space.

### DIFF
--- a/src/styles/theme/oc-button.scss
+++ b/src/styles/theme/oc-button.scss
@@ -28,7 +28,7 @@
 
   text-transform: none !important;
 
-  &:not(.uk-button-small):not(.uk-button-large) {
+  &:not(.uk-button-small):not(.uk-button-large):not(.uk-button-raw) {
     min-height: $global-control-height;
   }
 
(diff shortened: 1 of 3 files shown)


PVince81:
Fixes #654

src/styles/theme/oc-button.scss
```diff
@@ -28,7 +28,7 @@
 
   text-transform: none !important;
 
-  &:not(.uk-button-small):not(.uk-button-large) {
+  &:not(.uk-button-small):not(.uk-button-large):not(.uk-button-raw) {
```
kulmann: Since it is possible to set a size value and a variation separately, you should add `:not(.uk-button-raw)` to the other two classes (`.uk-button-small` and `.uk-button-large`) as well. I know that it's stupid, but just to be sure.

PVince81: done. please check my alternative solution

kulmann: Nice, even better!

PR #660, opened 2020-02-26
Make scrollbar styles more consistent

PVince81:
The Chrome scrollbar is now thinner to look like the Firefox one.

Adjusted Files page to be able to demonstrate the scrollbar styles.

Note: I heard that this would only affect Windows and Linux as Macs already have the proper style.

On my Linux box now the scrollbars look like this:

### Chromium
image

### Firefox
image

Fixes https://github.com/owncloud/phoenix/issues/3045

### DIFF
--- a/changelog/unreleased/660
+++ b/changelog/unreleased/660
@@ -1,0 +1,5 @@
+Bugfix: Made scrollbar styles consistent
+
+Scrollbar styles are now more consistent between Chrome and Firefox.
+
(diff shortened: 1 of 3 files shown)


PVince81:
Before this fix, in Chromium:
image

LukasHirt:
> Note: I heard that this would only affect Windows and Linux as Macs already have the proper style

It's affecting mac as well. In default, the scrollbar is appearing only when scrolling. This makes it visible permanently.

Issue #689, opened 2020-03-24
FR: message properties in oc-text-input


kulmann:
We need message properties (at least for errors, but might be useful for warnings as well) in `oc-text-input`, in order to
- show a (warning|error) state in the input element
- show a (warning|error) message below the input element
in a standardized way.

Also we want to provide an option (property) to have a fixed vertical space reserved below the input, so that a (warning|error) message doesn't break the layout below the input. E.g. we are using the input element in modals. An appearing error message pushes the buttons of the modal a few pixels to the bottom, which can be quite annoying.

Issue #688, opened 2020-03-24
Bug: oc-text-input is not shown in the generated docs


kulmann:
`oc-text-input` is not shown in the generated docs. Reason is, that the code in `@focus` produces a compilation warning.

PR #690, opened 2020-03-24
Introduce error and warning message below input

kulmann:
This PR introduces properties for the `oc-text-input` component, allowing to provide an error message or warning message, that will then be shown below the input field. In addition to that, the space below the input field can be set to a min height, so that the space is always reserved (at least for one line) no matter if an error or warning message is currently visible or not. Property is off by default, so that the input doesn't take up more space than it would have needed in the first place.
Issue is: https://github.com/owncloud/owncloud-design-system/issues/689

Also I noticed that the component was not present in the generated docs. This was due to a compilation warning in `@focus` of the input. Fixed it by moving the two commands into the method instead of having two commands inline in `@focus`.
Issue is: https://github.com/owncloud/owncloud-design-system/issues/688

### Error message
Bildschirmfoto 2020-03-24 um 17 26 40

### Warning message
Bildschirmfoto 2020-03-24 um 17 27 01

### No message
Bildschirmfoto 2020-03-24 um 17 26 52
You can notice that the upper input has a fixed vertical space below the input, so although no message is there, the vertical distance to the next input stays the same. The lower input doesn't have fixed vertical space below, so the space to the dark blue area shrinks when there is no message.

### DIFF
--- a/changelog/unreleased/688
+++ b/changelog/unreleased/688
@@ -1,0 +1,6 @@
+Bugfix: Fixed oc-text-input not appearing in generated docs
+
+The oc-text-input element didn't appear in the generated docs because of a compilation warning.
+
+https://github.com/owncloud/owncloud-design-system/issues/688
+https://github.com/owncloud/owncloud-design-system/pull/690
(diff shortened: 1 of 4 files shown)


LukasHirt:
Squash pls and can be merged 👍

kulmann:
Squashed 👍

PR #702, opened 2020-04-02
externalize uikit, vue-datetime, vue-avatar and moment

DeepDiver1975:
Trying to reduce the bundle size by:
- [x] externalizing some deps
- [ ] removing lodash


kulmann:
You might want to use [webpack-node-externals](https://www.npmjs.com/package/webpack-node-externals)

DeepDiver1975:
Oh. THX. Will Look into this.

PR #704, opened 2020-04-03
Fix wrapped text input

kulmann:
Revert `oc-text-input` back to having the input as root element, but with the small addition of defining input border colors (validation) through properties.

Extracted the validation message line and properties into a new `oc-form-input` pattern component, which then reuses `oc-text-input` as the input element.

### DIFF
--- a/src/patterns/OcFormInput.vue
+++ b/src/patterns/OcFormInput.vue
@@ -1,0 +1,191 @@
+<template>
+  <div>
+    <oc-text-input
+      v-bind="$_ocFormInput_inputBindings"
+      v-on="$listeners"
+      :has-error="!!errorMessage"
+      :has-warning="!!warningMessage"
+    />
+    <div class="oc-form-input-message" v-if="$_ocFormInput_showMessageLine">
+      <span v-if="!!warningMessage" class="oc-form-input-warning">{{ warningMessage }}</span>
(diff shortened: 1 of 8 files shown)


src/patterns/OcFormInput.vue
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * ## Accessibility
+ * The attributes `placeholder` and `aria-label` have different functions. The first specifies a short hint describing the expected value of an input field/text area, or gives an example (e.g. email@example.com). `aria-label` provides the accessible name of the text input (e.g. "Your address", "Comment",...).
+ */
+export default {
+  name: "oc-form-input",
+  components: { OcTextInput },
+  status: "review",
+  release: "1.0.0",
+  inheritAttrs: false,
```
kulmann, line 35: FYI this will be enhanced in vue 3, so that `$attrs` will include classes and styles as well. This way you can pass down attributes, classes, styles and listeners into a child component. `v-bind` on the `oc-text-input` in line 4 passes $attrs and $props into the text input. Neat feature.

kulmann: `inheritAttrs: false` has the effect, that $attrs and $listeners are not applied to the root element of the template, but on whichever other element you decide instead.

kulmann:
Not needed anymore since we solved this with https://github.com/owncloud/owncloud-design-system/pull/706

PR #709, opened 2020-04-07
Run unit tests in drone

LukasHirt:
Run unit tests on every PR, implemented example unit test for OCAlert component, add build steps to testing pipeline and run build pipeline only on tagging.


LukasHirt:
@micbar Do you think it's okay to run the build pipeline only on tags? Before there were two build steps which were run also on PRs but since we have them now also in the testing pipeline I think we could run it only on tags. I changed it for now in the drone.star but if you feel like it still should run I'll bring it back :wink:

.eslintignore
```diff
@@ -1,3 +1,4 @@
 node_modules/*
 dist/*
 test/unit/coverage/*
+src/**/*.spec.js
```
LukasHirt, line 4: We need to set our set of rules and move the eslint config to own file. I want to leave that to own PR so for now skipping eslint in tests since it would be failing.

LukasHirt: PR here #717

LukasHirt:
Unit tests are currently running under `yarn unit` because `yarn test` contains also build of design tokens. This is not needed though since the tokens are built during the build. Shall I remove it then from the `test` script and use there only the unit tests or should I leave it there? @PVince81 @kulmann

PVince81:
> Unit tests are currently running under `yarn unit` because `yarn test` contains also build of design tokens. This is not needed though since the tokens are built during the build of both docs and system. Shall I remove it then from the `test` script and use there only the unit tests or should I leave it there? @PVince81 @kulmann

if we have no purpose for `yarn test` then I guess it's ok to replace it

micbar:
@LukasHirt my experience with Ci tells me, we should run all steps on each PR.

Imagine the following scenario:

You break something in a PR, let it be docs or real code which only shows up in the build process.
Then you tag, and the build will fail.

You cannot correct the tag then and you would be completely stuck.

IMO, bring these steps back.

LukasHirt:
> my experience with Ci tells me, we should run all steps on each PR.

They are still being run as part of the testing pipeline. @micbar

https://github.com/owncloud/owncloud-design-system/pull/709/files#diff-74dc60589e0610c07e491b31f8693b2bR33
https://github.com/owncloud/owncloud-design-system/pull/709/files#diff-74dc60589e0610c07e491b31f8693b2bR42

PR #723, opened 2020-04-21
Added modal component

LukasHirt:
## Screenshot
image

## Open tasks:
- [x] Use variables for colours and gutters
- [x] Unit tests


src/elements/OcIcon.vue
```diff
@@ -17,7 +17,7 @@
 </template>
 
 <script>
-const req = require.context("@/assets/icons/", true, /^\.\/.*\.svg$/)
+const req = require.context("../assets/icons/", true, /^\.\/.*\.svg$/)
```
LukasHirt, line 20: Seems that the require-context-hook plugin is not resolving aliases.

LukasHirt:
For now, I put into examples also modal with input to show that we'll be able to use it in that way as well. However, as soon as we get rid of this use case in ocis web app, I'd drop this example not to encourage the devs to follow that path.

PR #741, opened 2020-05-04
Fixed prop mutation, wrong event and confirm button protection in the modal

LukasHirt:
Properly handle input value via v-model, use `input` event and disable confirm button in case of an error.

### DIFF
--- a/src/elements/OcModal.vue
+++ b/src/elements/OcModal.vue
@@ -22,13 +22,13 @@
           key="modal-input"
           ref="ocModalInput"
           class="oc-modal-body-input"
-          v-model="inputValue"
+          v-model="$_ocModal_input_value"
           :error-message="inputError"
           :label="inputLabel"
           :placeholder="inputPlaceholder"
           :disabled="inputDisabled"
           :fix-message-line="true"
-          @keydown="$_ocModal_input_type"
+          @input="$_ocModal_input_onInput"
           @change="$_ocModal_confirm"
         />
         <p
@@ -46,7 +46,7 @@
           <oc-button
             class="oc-modal-body-actions-confirm"
             :variation="$_ocModal_buttonConfirm_variation"
-            :disabled="buttonConfirmDisabled"
+            :disabled="buttonConfirmDisabled || !!inputError"
             v-text="buttonConfirmText"
             @click="$_ocModal_confirm"
           />
@@ -92,7 +92,7 @@ export default {
       required: false,
       default: "info",
       validator: (variation) => {
-        return variation.indexOf("info") > -1 || variation.indexOf("danger") > -1
+        return variation === "info" || variation === "danger"
       }
     },
     /**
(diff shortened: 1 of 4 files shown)


src/elements/OcModal.vue
```diff
@@ -213,23 +226,34 @@ export default {
...
        * 
-       * @property {String} inputValue Value of the input
+       * @property {String} value Value of the input
        */
-      this.$emit("type", this.$refs.ocModalInput.value)
+      this.$emit("type", value)
+    },
+
+    $_ocModal_input_assignPropAsValue() {
+      if (this.inputValue) {
```
kulmann, line 254: Just thinking... is the `if` necessary? 🤔 What's the reason to disallow setting empty or null values on `this.$_ocModal_input_value`?

LukasHirt: My original thinking was not to do any assignments in case the prop has not been used since it's not a required one. But you're right, this would prevent using the prop to e.g. clear the input. Thinking now however if it makes sense then to check if `hasInput === true` not to assign the value even though the input is missing.

src/elements/OcModal.vue
```diff
@@ -213,23 +226,32 @@ export default {
...
+       * @property {String} value Value of the input
        */
-      this.$emit("confirm", value)
+      this.$emit("confirm", this.$_ocModal_input_value)
     },
 
-    $_ocModal_input_type() {
+    $_ocModal_input_type(value) {
+      // Assign value to the input value
+      this.$_ocModal_input_value = value
```
kulmann, line 243: This is already handled through `v-model` on the input

src/elements/OcModal.vue
```diff
@@ -213,23 +226,32 @@ export default {
...
        * 
-       * @property {String} inputValue Value of the input
+       * @property {String} value Value of the input
        */
-      this.$emit("type", this.$refs.ocModalInput.value)
+      this.$emit("type", value)
+    },
+
+    $_ocModal_input_assignPropAsValue(value) {
+      this.$_ocModal_input_value = value
```
kulmann, line 254: Like above, this is already handled through `v-model` on the input. So you could get rid of this handler.

kulmann:
Why not just emit `input` instead of `type`? I think the `input` event reflects properly what is happening. What do you think? 🤔

LukasHirt:
> One more thing: I don't see an initializer for $_ocModal_input_value variable. You need to set that in the mounted lifecycle step, right?

This is done via the watcher.

> Why not just emit input instead of type? I think the input event reflects properly what is happening. What do you think? 🤔

Makes sense, will do 😉

src/elements/OcModal.vue
LukasHirt: This is not getting the value from the input. This assigns the prop called `inputValue` to the input.

LukasHirt:
@kulmann Ready for next review 😉

PR #747, opened 2020-05-08
Added coverage report step

LukasHirt:


LukasHirt:
@micbar I think this should be able to report the coverage to codacy. Can you pls check if the code makes sense and also add the token as a secret?

.drone.star
```diff
@@ -56,6 +56,20 @@ def testing(ctx):
                     'yarn test',
                 ]
             },
+            {
+                'name': 'report coverage to codacy',
+                'image': 'ubuntu',
+                'pull': 'always',
```
micbar, line 62: ubuntu? why not owncloudci/alpine? much smaller

.drone.star
```diff
@@ -56,6 +56,20 @@ def testing(ctx):
...
+                'commands': [
+                    'export CODACY_PROJECT_TOKEN=$CODACY_TOKEN',
+                    '/bin/bash <(curl -Ls https://coverage.codacy.com/get.sh)',
+                ],
+                'environment': {
+                    'CODACY_TOKEN': {
+                        'from_secret': 'codacy_token',
```
micbar, line 69: did you try to set `CODACY_PROJECT_TOKEN` directly

.drone.star
```diff
@@ -56,6 +56,20 @@ def testing(ctx):
                     'yarn test',
                 ]
             },
+            {
+                'name': 'report coverage to codacy',
```
micbar, line 60: `codacy` would nbe sufficient, alignment with the other pipelines

.drone.star
LukasHirt: I thought the token shouldn't be directly exposed in the code. But if it's not a problem I'll do it like that 😉

micbar: just the variable name, the value is not accessible -> secrets

LukasHirt: So skipping the export command and have it only as env?

micbar: But we agree to hide the token value in the secrets.

micbar: is now in the secrets

micbar: we need to reset it after the successful test

LukasHirt:
Getting error: `Exception in thread "main" java.net.UnknownHostException: api.codacy.com` now 😞

micbar:
> Getting error: `Exception in thread "main" java.net.UnknownHostException: api.codacy.com` now 😞

You now reached the same stage than me 2 weeks ago. 😄

LukasHirt:
Adding command `echo "$(dig +short api.codacy.com | tail -n1) api.codacy.com" >> /etc/hosts` as described [here](https://github.com/codacy/codacy-coverage-reporter/blob/master/README.md#substratesegfaulthandler-caught-signal-11) seems to help with the error but there are now two new ones.

This, however, requires getting `bind-tools` package before that step which is ugly 🤢 Otherwise we are getting an error about missing `dig` command.

LukasHirt:
> Adding command echo "$(dig +short api.codacy.com | tail -n1) api.codacy.com" >> /etc/hosts as described here  seems to help with the error but there are now two new ones.

Ok, now it fails again with the same error. So it doesn't seem to help.

LukasHirt:
I tried to use now the official [alpine image](https://hub.docker.com/_/alpine) and the result seems to be the same @micbar

micbar:
@LukasHirt got an answer in the ticket at codacy. Now using an alpine image with glibc as an intermediate solution.

LukasHirt:
> got an answer in the ticket at codacy. Now using an alpine image with glibc as an intermediate solution.

Awesome! Let's squash and merge then? @micbar

PR #768, opened 2020-05-25
Removed change and keydown events from input listeners

LukasHirt:
Since they are passed via listeners there is no need for them.

### DIFF
--- a/changelog/unreleased/remove-input-custom-events
+++ b/changelog/unreleased/remove-input-custom-events
@@ -1,0 +1,7 @@
+Change: Removed change and keydown events from text input
+
+We've removed change and keydown custom events from text input component.
+All listeners are passed to the input element so all events are still accessible.
+Focus and input events are still implemented as custom events.
+
+https://github.com/owncloud/owncloud-design-system/pull/768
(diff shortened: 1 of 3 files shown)


kulmann:
The `change` event actually exists for listening to changes, so I don't agree with removing it. There might be use cases for listening to all changes, not only submitted changes.

As we are passing all incoming listeners on our component through to the `input` component, we could get rid of all our custom event listeners. Thoughts?

LukasHirt:
> As we are passing all incoming listeners on our component through to the input component, we could get rid of all our custom event listeners. Thoughts?

That would make sense. I'm just thinking now if we should keep the `change` custom event and just rename it (just to make it a tiny bit easier for using) or also remove it and leave it to the place where it's used to handle the enter via `keydown` event.

kulmann:
> That would make sense. I'm just thinking now if we should keep the `change` custom event and just rename it (just to make it a tiny bit easier for using) or also remove it and leave it to the place where it's used to handle the enter via `keydown` event. @kulmann

I think using `@keydown.enter=...` already is fairly easy to use as a developer. I would remove the custom emitted change event entirely. One thing that might make sense is forcing the focus.

LukasHirt:
@kulmann I left there still `focus` and `input` events. `focus` additionally handles selecting of the input value and without `input` the `v-model` would be returning event instead of the value.

PR #770, opened 2020-05-25
Fixed href selectors in styles

LukasHirt:
There were a couple of selectors using href in the docs styles. They weren't escaping spaces though. This fixes it. Later I'd like to dig deeper to not include those sections in the sidebar but still include them in the final bundle but for now, I'd see this as sufficient solution.

### DIFF
--- a/docs/docs.styles.scss
+++ b/docs/docs.styles.scss
@@ -667,17 +667,17 @@ div[class^="rsg--toolbar"] {
 // Hide private components
 .hide-private,
 .hide-private + div,
-a[href="#private-components"],
-a[href="#private-components"] + ul {
+a[href="/#/Private%20Components"],
+a[href="/#/Private%20Components"] + ul {
   display: none !important;
 }
 
 // Hide Private components and placeholder sections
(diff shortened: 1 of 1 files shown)


docs/docs.styles.scss
```diff
@@ -667,17 +667,17 @@ div[class^="rsg--toolbar"] {
 // Hide private components
 .hide-private,
 .hide-private + div,
-a[href="#private-components"],
-a[href="#private-components"] + ul {
+a[href="/#/Private%20Components"],
+a[href="/#/Private%20Components"] + ul {
```
kulmann, line 671: Was it by accident to change this and the line above? Old version was an anchor on `#private-components`, now it's an own url.

LukasHirt: It's on purpose. As soon as we take care of the parent Private components item in the sidebar we do not need to care about any subitem which would be an anchor. Also, there is not such an anchor at all so I am not sure if it wasn't a bug or maybe it was different in an earlier version of styleguidist.

kulmann: Cool, thanks for clarifying!

PR #781, opened 2020-06-02
Automatically focus modal when mounted

LukasHirt:
## Description
When the modal is mounted, it receives automatically a focus. The focus is sent directly to the modal itself so skipping the wrapping div which works only as a background.

Refs https://github.com/owncloud/phoenix/issues/2263

### DIFF
--- a/src/elements/OcModal.vue
+++ b/src/elements/OcModal.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="oc-modal-background">
-    <div :class="$_ocModal_classes">
+    <div :class="$_ocModal_classes" tabindex="0" ref="$_ocModal">
       <div class="oc-modal-title">
         <oc-icon
           v-if="icon"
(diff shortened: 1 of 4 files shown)


PVince81:
focus is the first step, you should also automatically select whatever is inside the field already to make it easy to start typing

src/elements/OcModal.vue
```diff
@@ -217,6 +217,12 @@ export default {
     }
   },
 
+  mounted() {
+    this.$nextTick(() => {
+      this.$refs.$_ocModal.focus()
```
PVince81, line 222: does this focus on the input field ?

would it also focus on the cancel button if there is no input field ? (if that makes sense)

LukasHirt: Nope, it focuses the modal itself because of accessibility as described in here https://github.com/owncloud/phoenix/issues/2263#issuecomment-543516672

LukasHirt:
> focus is the first step, you should also automatically select whatever is inside the field already to make it easy to start typing

As soon as the input is focused (both click and keyboard) it already selects the whole value of it.

PR #804, opened 2020-06-30
Add mainContent slot into sidebar

LukasHirt:
The navigation inside of the sidebar can be replaced with a custom content via the slot called `mainContent`. In this slot can be included e.g. a short description of the product, guide the user through the current action, etc.

### DIFF
--- a/src/elements/OcSidebar.vue
+++ b/src/elements/OcSidebar.vue
@@ -24,7 +24,7 @@
         <!-- @slot Content above the navigation block -->
         <slot name="upperContent" />
       </div>
-      <nav :class="[{ 'uk-margin-bottom' : $slots.footer },  'oc-sidebar-nav']">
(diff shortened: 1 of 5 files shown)


src/elements/OcSidebar.vue
```diff
@@ -84,7 +87,7 @@ export default {
      */
     navItems: {
       type: Array,
-      required: true
+      required: false
```
kulmann, line 90: Please add an empty array as default. Otherwise there would be an error when `navItem` is not provided and the `maintContent` slot is not used.
like this: `default: () => []`

LukasHirt: Done

LukasHirt:
@kulmann I've already also squashed the commits since it's only one line change and if I'd do it afterwards the review would get dismissed automatically.

PR #858, opened 2020-09-15
Add new icons

LukasHirt:
We've added new icons which can be used to symbolise adding new shares, creating links and shared with lists.
We've also removed a wrong fill color from visibility off icon.

https://fontawesome.com/icons/share-square?style=solid
https://material.io/resources/icons/?search=group&icon=group_add&style=baseline

image


PVince81 (approved):
They look nice :+1:

PR #859, opened 2020-09-15
Introduce property for hiding the navigation in the sidebar

kulmann:
We introduced a new property for the sidebar for hiding the navigation
entirely. The internal logic of not showing the navigation under certain
circumstances has been removed.

Please note that this is a breaking change. The navigation inside the sidebar might become visible in situations where it was not visible before.


LukasHirt (approved):
LGTM 👍

PR #902, opened 2020-10-15
[UI] Removed explicit color from office document icons

rpocklin:
**Why?**
I discovered when changing some theme colours in phoenix that these icon colours, which are hard-coded into the SVG, depending on the colours you alter for the theme they contrast quite strongly and aren't easy to customise.

Removing the fill allows them to inherit the same colour as the sidebar which appears pleasing and is alter using SCSS.

**Before**
image

**After**
image


kulmann (approved):
Nice, thanks for you contribution! 🚀

PR #924, opened 2020-11-03
Add color prop to buttons

LukasHirt:


src/elements/OcButton.vue
```diff
@@ -226,4 +244,16 @@ Every button has to have an accessible name. It cannot be provided by a text bet
     <oc-button variation="danger">Delete</oc-button>
   </div>
 ```
+
+### Coloring buttons with variation "raw"
```
kulmann, line 248: Could you add a `default` one as well here, so that all existing options are visible in one place?

kulmann (approved):
Unit tests are failing. And please add default variant to the example in the docs. Other than that: perfect! :-)

PR #945, opened 2020-12-02
Remove toggle fullscreen button

LukasHirt:
The "toggle fullscreen button" in docs was useful when we were still having "pages" section where was a content spanning the whole pages. Since they are removed I do not see a point in having this feature anymore. This cleans a bit the examples view.

Screenshot 2020-12-02 at 10 15 06


kulmann (approved):
Yes, good idea! 👍

PR #964, opened 2020-12-04
Fix dev server and update vue styleguidist and template compiler

LukasHirt:
Dev server was broken so I fixed the command and config and at the same time updated two more deps. Will surpass #958 and #960 which are blocked by hanging CI.


kulmann (changes requested):
Local docs are not accessible when running `yarn dev`. 🤷  no error message in the terminal though.

LukasHirt:
> Local docs are not accessible when running yarn dev. 🤷 no error message in the terminal though.

Also if you run `yarn start`?

kulmann (approved):
Ah, yes, right. The URL is printed in the terminal, that's why I didn't give it another thought. ;-)

PR #975, opened 2020-12-06
Add new icons and fix router-link

LukasHirt:
We've added a `key` and `checklist` icons.
We've added an optional `to` prop to the `oc-tag` component so that it is properly assigned to the element in case of `router-link` type.


src/components/OcTag.vue
```diff
@@ -19,6 +19,15 @@ export default {
       required: false,
       default: "span",
       validator: (type) => type.match(/(span|button|router-link|a)/)
+    },
+
+    /**
+     * Target of the router link
+     */
+    to: {
```
kulmann, line 27: Why is it needed, when it is added to the root component anyway?

LukasHirt: TBH I am not sure but if it's not passed, then in ocis-web it doesn't render with console errors about missing required `to` prop and the whole path is then undefined :shug:

changelog/unreleased/new-icons
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add new icons
+
+We've added a `key` and `checklist` icons.
+
+https://fontawesome.com/icons/key?style=solid
+https://fontawesome.com/icons/tasks?style=solid
+https://github.com/owncloud/owncloud-design-system/pull/975
```
kulmann, line 7: The first link should always be the primary link. Could you move the PR link to the top?

src/components/OcTag.vue
kulmann: Ok 🤷 😅

changelog/unreleased/new-icons
LukasHirt: Done

kulmann (approved):
🚀

PR #997, opened 2020-12-11
fix non floating home icon in breadcrumb

fschade:
fix: https://github.com/owncloud/ocis/issues/1053


kulmann (approved):
Sign the CLA, then it's good to go 🚀

PR #1045, opened 2021-01-15
Add breakpoints tokens and classes

LukasHirt:
We've added tokens and utility classes to dynamically display/hide content depending on the breakpoints.


src/styles/theme/oc-visibility.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/* Desktop and smaller */
+@media (max-width: $breakpoint-large-max) {
+  .oc-visible\@xl {
+    display: none !important;
+  }
+}
+
+/*
```
kulmann, line 71: You could remove the comment. It still belongs to the `Visible` section

kulmann (approved):
👍

PR #1048, opened 2021-01-18
Remove lint pre-commit hook

LukasHirt:
As it's pretty standard that linting is happening in our IDEs and it's a first step we run in tests which on failure fails the whole test, we want to drop this in order to not run some hooks which don't bring much benefits.


kulmann (approved):
👍

Issue #1049, opened 2021-01-18
Huge system.css file on build


kulmann:
With the [current master](https://github.com/owncloud/owncloud-design-system/tree/fde807b5a72087d6f7eda784925513928a9a583e) the `system.css` file is 7MB big, compared to just ~400KB in [an older state](https://github.com/owncloud/owncloud-design-system/tree/c6ff272fd9f8f3d343da563821068445006f90db).

kulmann:
Unfortunately we squashed ~55 commits of the branch with the new oc-table and oc-table-files components into just one commit. Went back in time with `git reflog` to find:
- a bad commit before the squash: `cfe66ec7060cbda7ba406e3059a9a9d2e1a4d8d9 `
- and a good commit before the squash: `c6ff272fd9f8f3d343da563821068445006f90db`

Bisecting now.

kulmann:
Bisecting just brought the hunch that with every introduced component the `system.css` got bigger - apparently because we defined component styles as `scoped`. Removing all `scoped` attributes from the `<style>` definitions brings the size of `system.css` back to the (still huge) 403KB.

PR #1050, opened 2021-01-18
Fix style sizes

kulmann:
Needed to get rid of the `scoped` props. This blew up the css bundle size to 7 MB instead of 400 KB.

Fixes https://github.com/owncloud/owncloud-design-system/issues/1049


kulmann:
Also, CI time is back to 3min 20sec with this PR (coming from 4min 15sec).

LukasHirt (approved):
Do you know why the `scoped` styles actually caused the bump in the size? Feels super strange that this would have such a side-effect.

kulmann:
> Do you know why the `scoped` styles actually caused the bump in the size? Feels super strange that this would have such a side-effect.

Somehow scoped styles are cloning **all** styles (not just the ones in the scope of the component).
Looks like this in the css file: `.oc-accordion-title[data-v-33c90a04]`.
This looks like a webpack configuration issue, but thinking about it I think that the styles should not be scoped to give developers a chance to re-use the classes. Or do we explicitly not want them to be reusable?

LukasHirt:
> styles should not be scoped to give developers a chance to re-use the classes. Or do we explicitly not want them to be reusable?

Hmm, as we have them tied closely to the component, is there actually a use case for using them without those components? I'd prefer to always use the components. But if there's some use case I am overlooking then sure, let's have them re-usable. Of course, an exception would be some utility classes but those shouldn't live in a component.

I was actually super happy that the styles would be scoped and wouldn't have a chance to fight with any other classes in consuming apps.

kulmann:
Hmmmmm... true, true.
I'll create an issue that we want to solve it to have component styles scoped again.

PR #1055, opened 2021-01-19
Add pagination component

LukasHirt:
image

### DIFF
--- a/src/components/pagination/OcPagination.vue
+++ b/src/components/pagination/OcPagination.vue
@@ -1,0 +1,240 @@
+<template>
+  <nav class="oc-pagination" :aria-label="$gettext('Pagination')">
+    <ul class="oc-pagination-list">
+      <li v-if="isPrevPageAvailable" class="oc-pagination-list-item">
+        <router-link
+          class="oc-pagination-list-item-prev"
+          :aria-label="$gettext('Go to the previous page')"
+          :to="previousPageLink"
+        >
+          <oc-icon name="chevron_left" />
+        </router-link>
+      </li>
+      <li v-for="(page, index) in displayedPages" :key="index" class="oc-pagination-list-item">
+        <component
+          :is="pageComponent(page)"
+          :class="pageClass(page)"
+          v-bind="pageProps(page)"
+          v-text="page"
+        />
+      </li>
+      <li v-if="isNextPageAvailable" class="oc-pagination-list-item">
+        <router-link
+          class="oc-pagination-list-item-next"
+          :aria-label="$gettext('Go to the next page')"
+          :to="nextPageLink"
+        >
+          <oc-icon name="chevron_right" />
+        </router-link>
+      </li>
+    </ul>
+  </nav>
+</template>
+
+<script>
+import OcIcon from "../OcIcon.vue"
+
+/**
+ * A list of links used for switching to different pages
+ */
+export default {
+  name: "OcPagination",
+
+  components: { OcIcon },
+
+  props: {
+    /**
+     * Number of pages to be displayed
+     */
+    pages: {
+      type: Number,
+      required: true,
+    },
+    /**
+     * Currently active page
+     */
+    currentPage: {
+      type: Number,
+      required: true,
+    },
+    /**
+     * Number of pages to be displayed. Needs to be an even number.
+     * Pages will be equaly split into two parts and remaining pages will be displayed as ellipsis
+     */
+    maxDisplayed: {
+      type: Number,
+      required: false,
+      default: null,
+      validator: value => {
+        if (value % 2 === 0) {
+          return true
+        }
+
+        // Since Vue doens't support custom validator error message, log the error manually
+        console.error("maxDisplayed needs to be an even number")
+      },
+    },
+    /**
+     * Current route which is used to render pages
+     */
+    currentRoute: {
+      type: String,
+      required: true,
+    },
+  },
+
+  computed: {
+    displayedPages() {
+      if (this.maxDisplayed && this.maxDisplayed < this.pages) {
+        let pages = []
+
+        for (let i = 0; i < this.pages; i++) {
+          pages.push(i + 1)
+        }
+
+        const pagesRight = pages.slice(this.currentPage, this.currentPage + this.maxDisplayed / 2)
+        const pagesLeft = pages.slice(
+          this.currentPage - 1 - this.maxDisplayed / 2,
+          this.currentPage - 1
+        )
+
+        pages = [...pagesLeft, this.currentPage, ...pagesRight]
+
+        pages[0] > 2 ? pages.unshift(1, "...") : pages.unshift(1)
+        pages[pages.length - 1] < this.pages - 1
+          ? pages.push("...", this.pages)
+          : pages.push(this.pages)
+
+        return pages
+      }
+
+      return this.pages
+    },
+
+    isPrevPageAvailable() {
+      return this.currentPage !== 1
+    },
+
+    isNextPageAvailable() {
+      return this.pages !== this.currentPage
+    },
+
+    _currentRoute() {
+      return JSON.parse(JSON.stringify(this.currentRoute)).replace(/\/$/, "")
+    },
+
+    previousPageLink() {
+      return this._currentRoute + "/" + (this.currentPage - 1)
+    },
+
+    nextPageLink() {
+      return this._currentRoute + "/" + (this.currentPage + 1)
+    },
+  },
+
+  methods: {
+    pageLabel(page) {
+      const translated = this.$gettext("Go to page %{ page }")
+
+      return this.$gettextInterpolate(translated, { page })
+    },
+
+    isCurrentPage(page) {
+      return this.currentPage === page
+    },
+
+    pageComponent(page) {
+      return page === "..." || this.isCurrentPage(page) ? "span" : "router-link"
+    },
+
+    pageProps(page) {
+      if (this.isCurrentPage(page)) {
+        return {
+          "aria-current": "true",
+        }
+      }
+
+      return {
+        "aria-label": this.pageLabel(page),
+        to: this._currentRoute + "/" + page,
+      }
+    },
+
+    pageClass(page) {
+      const classes = ["oc-pagination-list-item-page"]
+
+      if (this.isCurrentPage(page)) {
+        classes.push("oc-pagination-list-item-current")
+      } else if (page === "...") {
+        classes.push("oc-pagination-list-item-ellipsis")
+      } else {
+        classes.push("oc-pagination-list-item-link")
+      }
+
+      return classes
+    },
+  },
+}
+</script>
+
+<style lang="scss">
+.oc-pagination {
+  &-list {
+    align-items: center;
+    display: flex;
+    gap: var(--oc-space-small);
+    list-style: none;
+
+    &-item {
+      &-page {
+        border-radius: 4px;
+        color: var(--oc-color-swatch-brand-default);
+        padding: var(--oc-space-xsmall) var(--oc-space-small);
+        transition: background-color $transition-duration-short ease-in-out;
+
+        &:not(span):hover,
+        &:not(span):focus {
+          background-color: var(--oc-color-swatch-brand-default);
+          color: var(--oc-color-text-inverse);
+          text-decoration: none;
+        }
+      }
+
+      &-current {
+        background-color: var(--oc-color-swatch-brand-default);
+        color: var(--oc-color-text-inverse);
+        font-weight: bold;
+      }
+
+      &-prev,
+      &-next {
+        display: flex;
+
+        > .oc-icon > svg {
+          fill: var(--oc-color-swatch-brand-default);
+        }
+      }
+    }
+  }
+}
+</style>
+
+<docs>
+## Examples
+```vue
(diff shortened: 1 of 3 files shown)


pascalwengerter:
Superseded by #1320

LukasHirt:
> Superseded by #1320

Nope, this PR introduces the pagination itself. Mentioned PR takes care of bringing it into the files table

LukasHirt:
> Uff, sorry for the misunderstanding 🥴

No worries 😉

kulmann:
> > Superseded by #1320
> 
> Nope, this PR introduces the pagination itself. Mentioned PR takes care of bringing it into the files table

Didn't we decide recently that we want to bring the pagination into the views inside web? 🤔 Basic idea was: when we implement a grid view as well, we'll need components usable in web for
- sorting
- pagination
- a switch between grid-view and list-view

and then depending on the chosen view (grid or list) only the component for displaying files needs to be mounted in the web views, while the sorting, pagination and view-mode-switch can stay the same.

LukasHirt:
> Didn't we decide recently that we want to bring the pagination into the views inside web?

Yep, and this component is what should be used there. Sorry for the confusion with the follow-up PR. Didn't want to mention it here as it felt like a concern of the other PR and there it's already written I think in comments.

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
+<template>
+  <nav class="oc-pagination" :aria-label="$gettext('Pagination')">
+    <ul class="oc-pagination-list">
```
kulmann, line 3: Should be an ordered list `<ol>`, right? 🤔

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+          : pages.push(this.pages)
+
+        return pages
+      }
+
+      return this.pages
+    },
+
+    isPrevPageAvailable() {
+      return this.currentPage !== 1
```
kulmann, line 141:
```suggestion
      return this.currentPage > 1
```
to make it more resilient (e.g. when page number comes from the url)

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+    },
+
+    isNextPageAvailable() {
+      return this.pages !== this.currentPage
```
kulmann, line 145:
```suggestion
      return this.pages > this.currentPage
```

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+    },
+
+    pageComponent(page) {
+      return page === "..." || this.isCurrentPage(page) ? "span" : "router-link"
+    },
+
+    pageProps(page) {
+      if (this.isCurrentPage(page)) {
+        return {
+          "aria-current": "true",
```
kulmann, line 206:
```suggestion
          "aria-current": "page",
```

src/components/pagination/OcPagination.vue
kulmann: to make it more resilient (e.g. when page number comes from the url)

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+    isNextPageAvailable() {
+      return this.pages !== this.currentPage
+    },
+
+    _currentRoute() {
+      return JSON.parse(JSON.stringify(this.currentRoute)).replace(/\/$/, "")
+    },
+
+    previousPageLink() {
+      return this._currentRoute + "/" + (this.currentPage - 1)
```
kulmann, line 153: Can we have a `targetRoute` object instead and replace a `page` param within the params of that targetRoute object? That would make the route handling here more versatile for different route formats. This would already be consistent with what we're doing for navigation in the files table (expecting a route object and altering it).

LukasHirt:
Found a bug now while testing in web... somehow first page is being doubled sometimes

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+      } else if (page === "...") {
+        classes.push("oc-pagination-list-item-ellipsis")
+      } else {
+        classes.push("oc-pagination-list-item-link")
+      }
+
+      return classes
+    },
+
+    navigateToPage() {
```
kulmann, line 230: Probably  a leftover from development

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+      return page === "..." || this.isCurrentPage(page) ? "span" : "router-link"
+    },
+
+    pageProps(page) {
+      if (this.isCurrentPage(page)) {
+        return {
+          "aria-current": "true",
+        }
+      }
+
```
kulmann, line 209: This needs another early return for `if (page === '...')`

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,337 @@
...
+    },
+    /**
+     * Currently active page
+     */
+    currentPage: {
+      type: Number,
+      required: true,
+    },
+    /**
+     * Number of pages to be displayed. Needs to be an even number.
```
kulmann, line 75: What's the reason for requiring an even number?

kulmann (changes requested):
Found some issues. Could you also introduce a computed prop for the `currentPage` which always makes sure that the current page is within bounds? E.g. when `pages` is 5 and `currentPage` is 10, then use `5` as internal value.

kulmann:
@tbsbdr could you have a look at the screenshot in the PR description please? I'm not exactly sure if I like the "go to" input but I'm also not sure about a different solution. Maybe we can use our `OcSelect` component. And some kind of separator (or just more margin) between the page navigation and the go to would be good as well I think.

src/components/pagination/OcPagination.vue
LukasHirt: That number is evenly split into both sides. If there's 4 max visible, 2 will be visible on each side of the current page. I wanted to have it originally as the number for those sides separately i.e. 2 = 2 on each side but I have gone with this way mainly due to not being able to come up with the correct naming.

LukasHirt:
> I'm not exactly sure if I like the "go to" input but I'm also not sure about a different solution.

I'm also not fully sold on the current UI but went with this as the most working solution for now. Few points from my thought process when implementing it:
- The label looks strange but having it on top of the input creates super weird vertical alignment
- Select might look nice but having to scroll in the select if there'd be too many pages would just be tedious like selecting a year of birth in a select is
- Already tried to give it a bit more space between the pagination and the input but there's little space to experiment here as the whole component is already really wide for mobile screens

kulmann:
Fully agree with your thoughts and the outcome @LukasHirt 

I'd propose to have the pagination component without any `go to` in the first iteration. We can think about that later on in a subsequent PR.
Thoughts? @tbsbdr @LukasHirt

kulmann:
@LukasHirt when I try keyboard navigation in the docs, I don't see any focus styles. Can you look into that?

tbsbdr:
I'm not sure, if we need the pagenumbers, because for the user, they are actually a "shot in the dark". For folder with very many files (2.000+), you are lost anyway i guess without search or filtering / sorting. sorting gets crucial if we have a pagination.

@LukasHirt what do you think of a more simplified approach?

## simple prev/next with "files per page"
image

## Show more
(maybe causes a11y problems? but would be my favourite)
image

src/components/pagination/OcPagination.vue
kulmann: Ah, makes sense. But then it should enforced to be an uneven number - you always have the current page and want to display the same amount of pages on both sides. Which makes `x+x+1 = uneven`.

kulmann:
@LukasHirt I just had a call with @tbsbdr to come to a decision how we want to have it in the first iteration. Outcome: the way you have it in your original screenshot, but without the `go to` section. So, really just simple pagination, nothing else. Additional buttons for jumping to the first and last page would make sense, if you can find icons for it. If not leave it out.

```
« < … 4 5 6 7 8 … > »
```
with 6 being the current page.

LukasHirt:
> I'm not sure, if we need the pagenumbers, because for the user, they are actually a "shot in the dark".

I think it's still nice to have the pages as you might create some "mind map" over time and remember that the file you're looking for is on page x. On the other hand.

> Outcome: the way you have it in your original screenshot, but without the go to section.

TBH honest I am not a fan of not giving the user a simple way of navigating to a specific page (see also opinion about navigation to pages above). Especially since we don't even have the search in oCIS at all. And having to tell users to "edit the URL" to navigate to a different page or just keep hitting the next button till you're there are neither good approaches IMHO.

> Additional buttons for jumping to the first and last page would make sense

I've added the always visible first and last page exactly for these jumps. I think it's nice that the user sees how many pages are there actually.

Don't get me wrong, I like very much the approach to simplify the pagination. It's just that I think we'd be getting rid of actually useful features.

kulmann:
> TBH honest I am not a fan of not giving the user a simple way of navigating to a specific page (see also opinion about navigation to pages above). Especially since we don't even have the search in oCIS at all. And having to tell users to "edit the URL" to navigate to a different page or just keep hitting the next button till you're there are neither good approaches IMHO.

But if we can already tell that the solution feels weird, I'd prefer to have the simpler approach quickly now, and then do another UX session on top of that with our well known early adopter to gather ideas what would work. ;-) Fully agree that editing the URL is not something we'd tell any user.

> I've added the always visible first and last page exactly for these jumps. I think it's nice that the user sees how many pages are there actually.

Thanks for pointing that out, didn't realize that before. Yes, I agree, always visible first and last page is a good solution, I'm happy with that. 👍

src/components/pagination/OcPagination.vue
LukasHirt: In my head, I took the current page outside of this number. Which doesn't make any sense with the current wording. What about changing the docs to `Number of pages to be displayed next to the current page evenly split into both sides. It is required to use an even number.`

kulmann: IMO creates too much complexity. All paginations that I ever used treated this as the total number of displayed pages, including the current page. 😉

kulmann: ... which is simple and doesn't require you to calculate `x - 1` in your head.

LukasHirt:
> when I try keyboard navigation in the docs, I don't see any focus styles. Can you look into that?

This is expected... the pagination links in the docs haven't got `href` attribute since they are just a mock of router-link. Anchor tag without `href` doesn't get focused.

LukasHirt:
@kulmann Ready for final review

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,277 @@
...
+    previousPageLink() {
+      return this.bindPageLink(this.$_currentPage - 1)
+    },
+
+    nextPageLink() {
+      return this.bindPageLink(this.$_currentPage + 1)
+    },
+
+    $_currentPage() {
+      return Math.min(this.currentPage, this.pages)
```
kulmann, line 138:
```suggestion
      return Math.max(1, Math.min(this.currentPage, this.pages))
```

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,277 @@
...
+    displayedPages() {
+      if (this.maxDisplayed && this.maxDisplayed < this.pages) {
+        let pages = []
+
+        for (let i = 0; i < this.pages; i++) {
+          pages.push(i + 1)
+        }
+
+        const currentPageIndex = this.$_currentPage - 1
+        const indentation = (this.maxDisplayed - 1) / 2
```
kulmann, line 98:
```suggestion
        const indentation = Math.floor(this.maxDisplayed / 2)
```
If we ever decide to also allow even values for `maxDisplayed` this would already work out of the box, i.e. code less error prone in the future. With odd number the result is the same.

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,277 @@
...
+        if (this.$_currentPage < this.pages - 1) {
+          pages[pages.length - 1] < this.pages - 1
+            ? pages.push("...", this.pages)
+            : pages.push(this.pages)
+        }
+
+        return pages
+      }
+
+      return this.pages
```
kulmann, line 118: The early return is an array, while this return here is an int. Needs to be an array as well.

src/components/pagination/OcPagination.vue
```diff
@@ -0,0 +1,277 @@
...
+      } else if (page === "...") {
+        classes.push("oc-pagination-list-item-ellipsis")
+      } else {
+        classes.push("oc-pagination-list-item-link")
+      }
+
+      return classes
+    },
+
+    bindPageLink(page) {
```
kulmann, line 190: 😍

kulmann (changes requested):
Getting close 🏃

src/components/pagination/OcPagination.vue
LukasHirt: Vue is smart enough to understand the number and use it as an array so I wanted to save the extra loop to create the array... but for consistency sake :+1:

kulmann (approved):
Awesome! 😍 🚀

PR #1064, opened 2021-01-21
Pin UIkit to v3.5.16

LukasHirt:
We've pinned the version of UIkit to 3.5.16.
The reason for this is that the version 3.5.17 caused a regression to our autocomplete component.
Since we want to remove UIkit from the ODS in the near future, we are not taking the time to come up with a fix.
We do not plan to upgrade UIkit anymore.


kulmann (approved):
👍

PR #1082, opened 2021-02-08
Remove URL icon prop

LukasHirt:
We've removed the URL prop from icons which was replacing the internal icons with an image loading the icon from arbitrary URLs.
As of now, only internal icons can be used in the `oc-icon` component.


kulmann (changes requested):
We have some places in oC Web where we used the `oc-icon` component with a url (NotFoundMessage, ApplicationsMenu, UserMenu). While those could be replaced with the `oc-img` component, there are three things to solve - either by documentation or by adding features to the `oc-img` component:

1) oc-img has no aria-label
2) oc-img has no variation classes
3) oc-img has no size classes

Maybe a proper solution would be to introduce an `oc-svg` component? Because thre props from above partially don't make much sense for an image component. If you want to tackle that in a separate PR then I'd be fine approving this one. But we would need it for an upgrade in oC Web anyway.

LukasHirt:
@kulmann What would you say about extending the icon set we have in case it would be needed and stick with the `oc-icon` even in those cases? That would mean also switching to `name` prop instead of `url` in the applications menu which would limit the icons to internal ones only but that's actually what we wanted to do by this change, isn't it?

kulmann:
Agreed. 👍 
But could you add the following svg to our icon set in this PR then, so that we have everything we need for switching over to `oc-icon`s with names in web directly? It's used in the files app for the 404. https://github.com/owncloud/web/blob/master/packages/web-container/themes/owncloud/assets/cloud.svg

pascalwengerter:
Rebased on master to reflect @JammingBen's most recent changes regarding SVG handling and adding the icon @kulmann requested 💪🏽

src/components/OcIcon.vue
```diff
@@ -238,15 +186,6 @@ export default {
       </oc-tbody>
     </oc-table-simple>
 
-    <h3 class="uk-heading-divider">
-      Icons loaded via URL
-    </h3>
-    <div class="oc-m">
-      <oc-icon size="medium" url="https://interactive-examples.mdn.mozilla.net/media/examples/firefox-logo.svg" accessible-label="Firefox logo"/>
-      <oc-icon size="large" url="https://interactive-examples.mdn.mozilla.net/media/examples/firefox-logo.svg"/>
-      <oc-icon size="large" name="account_circle" url="https://interactive-examples.mdn.mozilla.net/media/examples/firefox-logo.sv" accessible-label="Account"/>
```
kulmann, line 247: Could you keep this line but get rid of the url attribute? would be nice to keep an example showing the `accessible-label` property. + copy the same line and without accessible-label.

pascalwengerter: Sure!

JammingBen (changes requested):
LGTM overall 👍 

Now that we removed all `` tags in the component, we should also get rid of all the img-related stuff in the accessibility info: https://github.com/owncloud/owncloud-design-system/blob/master/src/components/OcIcon.vue#L47). It's just confusing otherwise.

Ping me if you need help or want me to adjust the text as I recently wrote it :)

pascalwengerter:
> LGTM overall 👍
> 
> Now that we removed all `` tags in the component, we should also get rid of all the img-related stuff in the accessibility info: https://github.com/owncloud/owncloud-design-system/blob/master/src/components/OcIcon.vue#L47). It's just confusing otherwise.
> 
> Ping me if you need help or want me to adjust the text as I recently wrote it :)

I tried updating the docs myself, happy to have you review it (and great catch)!

kulmann (approved):

JammingBen (approved):

PR #1084, opened 2021-02-09
Remove basic styles

LukasHirt:
We've removed styles from the body element and from the #oc-content element. We were forcing styles on a global level which were too specific and because of that limited the usage of the design system. Any specific styling like that should be done in the consuming app.


kulmann (approved):
full ack 👍

PR #1085, opened 2021-02-11
Merge resource target path with folder path

LukasHirt:
We've merged the resource target route path with the folder path in the `router-link`. When passing the path as a parameter, it wasn't being recognised and the navigation just pointed to the target route path. By explicitly merging them, we make sure that the navigation behaves as expected.

Additional changes in this PR:
- Encode the path
- Stop propagation for the links to prevent opening sidebar in oC Web

### DIFF
--- a/src/components/resource/OcResource.vue
+++ b/src/components/resource/OcResource.vue
@@ -15,6 +15,7 @@
         v-bind="componentProps"
         v-if="isResourceClickable"
         @click.stop="emitClick"
+        @click.native.stop
       >
         <oc-resource-name :full-path="resource.path" :is-path-displayed="isPathDisplayed" />
       </component>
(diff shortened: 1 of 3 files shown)


src/components/resource/OcResource.vue
```diff
@@ -99,6 +100,14 @@ export default {
       return this.isResourceClickable && this.isFolder
     },
 
+    folderLink() {
+      const path = this.resource.path.startsWith("/")
```
kulmann, line 104: These three lines could be a one liner with a regexp for stripping the leading slash only if it exists:
`const path = this.resource.path.replace(/^\//, ''); // remove leading slash`

kulmann (approved):
Added a comment for possible reduction of code complexity (IMO still readable, you can add the comment if you want to).

We already know that we want to build URLs for resources without the encoded path in the future. But since we are not there, yet, I agree with this PR! 👍

src/components/resource/OcResource.vue
LukasHirt: Thanks - adjusted :wink:

kulmann (approved):

PR #1091, opened 2021-02-16
Explicitly set oc-button:hover color & text-decoration

### LINKED ISSUE #899: Button Hover wrong colors
### Steps to reproduce
1. Hover over a primary button

### Expected behaviour
The text should still be visible.

### Actual behaviour
The text turns into a strange color and the text is not visible anymore. See screenshot.

Weird thing is that this doesn't happen in ODS, but only in phoenix.

Screenshot 2020-10-09 at 09 31 45

pascalwengerter:
Fixes #899 (but there's more to it, see details below)

I tested the changes locally editing the code preview on the [OcButton](http://localhost:6060/#/oC%20Components/OcButton), as

```html
  <oc-button variation="primary" class="oc-mr-s" type="a">Primary button (a)</oc-button>
  <oc-button variation="primary" class="oc-mr-s" type="button">Primary button (button)</oc-button>
```

Having resolved the `color` problem I realized that `text-decoration` differs between `type="a"` and `type="button"`, so I decided to explicitly set it to `none` (as this is the case with other button types in use).

Just now I made the mistake and did some further investigation since oc-buttons can also be `"danger"` or `"inverse"` (default), and there's more odd behaviour with `color`, `text-decoration` but also `font-family`? 🤯 

To reproduce (and open further issues), feel free to use the snippet below:

```html
<oc-button variation="danger" type="button" class="oc-mr-s">
  <oc-icon name="delete" aria-hidden="true" />
  Danger Button
</oc-button>
<oc-button variation="danger" type="a" class="oc-mr-s">
  <oc-icon name="delete" aria-hidden="true" />
  Danger Button
</oc-button>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,19 @@ The following sections list the changes in ownCloud Design System unreleased.
 
 ## Summary
 
+* Bugfix - Button hover color fix: [#1091](https://github.com/owncloud/owncloud-design-system/pull/1091)
 * Change - Merge resource target path with folder path: [#1085](https://github.com/owncloud/owncloud-design-system/pull/1085)
 * Change - Remove basic styles: [#1084](https://github.com/owncloud/owncloud-design-system/pull/1084)
 * Enhancement - Remove the trailing slash from resource target path: [#1087](https://github.com/owncloud/owncloud-design-system/pull/1087)
 
 ## Details
+
+* Bugfix - Button hover color fix: [#1091](https://github.com/owncloud/owncloud-design-system/pull/1091)
+
+   When hovering the oc-button (*primary*) component had different styling, depending on if it was used as `a` or `button` tag.
+   This was caused by the underlying `ui-kit` styles, since no specific component-based styling was provided.
+   By making the hover styling explicit in the component stylesheet, this is fixed.
+   https://github.com/owncloud/owncloud-design-system/pull/1091
 
 * Change - Merge resource target path with folder path: [#1085](https://github.com/owncloud/owncloud-design-system/pull/1085)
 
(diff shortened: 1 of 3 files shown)


LukasHirt:
Super weird that the font-family is changing with different variations. Anyway, thanks for taking care! Add a changelog item pls ([changelog readme](https://github.com/owncloud/owncloud-design-system/blob/master/changelog/README.md)) and I think we can merge this PR and take care of other issues in a separate one

kulmann (changes requested):
The changelog in our projects is autogenerated, based on individual changelog files. Could you please place your changelog item in the `changelogs/unreleased` folder? https://github.com/owncloud/owncloud-design-system/tree/master/changelog/unreleased

Format of one changelog file is super simple: Title, description, list of links. Each of those categories separated by linebreaks.

pascalwengerter:
> The changelog in our projects is autogenerated, based on individual changelog files. Could you please place your changelog item in the `changelogs/unreleased` folder? https://github.com/owncloud/owncloud-design-system/tree/master/changelog/unreleased
> 
> Format of one changelog file is super simple: Title, description, list of links. Each of those categories separated by linebreaks.

@kulmann hope this one checks out? Is there a clear rule on squashing commits etc? :)

kulmann (approved):
👍

kulmann:
> > The changelog in our projects is autogenerated, based on individual changelog files. Could you please place your changelog item in the `changelogs/unreleased` folder? https://github.com/owncloud/owncloud-design-system/tree/master/changelog/unreleased
> > Format of one changelog file is super simple: Title, description, list of links. Each of those categories separated by linebreaks.
> 
> @kulmann hope this one checks out? Is there a clear rule on squashing commits etc? :)

We try to avoid squashing if possible and have small commits with reasonable messages for improved bisecting experience and easy reviews etc. I don't mind having reverts in the commit log.

PR #1093, opened 2021-02-17
Add isSelected and disabled props

LukasHirt:
We've added `disabled` prop to `oc-table` and `oc-table-files` components.
This prop is used to pass IDs of disabled resources so that they are clearly distinguishable from the other resources.

We've added `isSelectable` prop to `oc-table-files` component.
This prop is used to assert whether resources in the table can be selected.


src/components/table/OcTableFiles.vue
```diff
@@ -316,6 +336,14 @@ export default {
        */
       this.$emit("fileClick", resource)
     },
+
+    $_isResourceClickable(resourceId) {
```
kulmann, line 340: Hm... didn't we say we want to skip the `$_` and namespacing as long as we're not in a mixin? 🤔  I get that you needed to have a different name than `isResourceClickable`, because that is the prop name already. Can we rename the prop to `areResourcesClickable`, since it is deciding it for all resources?

src/components/table/OcTableFiles.vue
```diff
@@ -316,6 +336,14 @@ export default {
...
+      const isDisabled = Array.isArray(this.disabled)
+        ? this.disabled.contains(resourceId)
+        : this.disabled === resourceId
+
+      return this.isResourceClickable && !isDisabled
```
kulmann, line 345: Would be good to have an early return on the prop. I.e. as a first check in the method: 
```
if (!this.isResourceClickable) {
   return false
}
```
And then 
```
return Array.isArray(this.disabled) ? this.disabled.contains(resourceId) : this.disabled === resourceId
```

kulmann (changes requested):

kulmann (approved):
👍

PR #1094, opened 2021-02-17
Use values as keys

LukasHirt:
We've started using values as keys for the table celss so that they are reactive to the changes in values.

### DIFF
--- a/changelog/unreleased/table-cells-keys
+++ b/changelog/unreleased/table-cells-keys
@@ -1,5 +1,5 @@
 Change: Use values as keys for table cells
 
-We've started using values as keys for the table celss so that they are reactive to the changes in values.
+We've started using values as keys for the table cells so that they are reactive to the changes in values.
 
 https://github.com/owncloud/owncloud-design-system/pull/1094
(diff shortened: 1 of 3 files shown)


changelog/unreleased/table-cells-keys
```diff
@@ -0,0 +1,5 @@
+Change: Use values as keys for table cells
+
+We've started using values as keys for the table celss so that they are reactive to the changes in values.
```
kulmann, line 3:
```suggestion
We've started using values as keys for the table cells so that they are reactive to the changes in values.
```

src/components/table/OcTable.vue
```diff
@@ -31,7 +31,7 @@
       >
         <oc-td
           v-for="field in fields"
-          :key="`oc-tbody-tr-${item[idKey] || index}-${field.name}`"
+          :key="cellKey(field, index, item)"
```
kulmann, line 34:
```suggestion
          :key="'oc-tbody-tr-' + cellKey(field, index, item)"
```
I would like to keep the key prefixes consistent with all the other keys in this component.

kulmann (changes requested):

kulmann (approved):

PR #1099, opened 2021-02-19
Add table footer

LukasHirt:
We've added a footer to the table and table files components. To add any content to the footer, use `footer` slot.


kulmann (changes requested):
Could you add a simple example for the docs, in both components?

kulmann (approved):
👍

PR #1100, opened 2021-02-19
Make dropzone public

LukasHirt:
We've made the dropzone component public so that it appears in the documentation. We've also moved all related styles from the dedicated file into the component and added a border.


src/components/OcDropzone.vue
```diff
@@ -0,0 +1,40 @@
...
+</template>
+<script>
+export default {
+  name: "OcDropzone",
+  status: "review",
+  release: "0.0.1",
+}
+</script>
+
+<style lang="scss" scoped>
```
kulmann, line 15:
```suggestion
<style lang="scss">
```
Could you drop the `scoped` for now, until we find out how to have scoped styles without blowing up the bundle size? :-) `system.css` size increased from 403K on master to 919K on this branch.

kulmann (changes requested):

src/components/OcDropzone.vue
LukasHirt: Done

kulmann (approved):

PR #1102, opened 2021-02-22
Pass route object instead of string in targetRoute

LukasHirt:
We've changed the `targetRoute` prop in the `oc-resource` component to accept object instead of a string. This object is now accepting keys `name`, `params` and `query`. Only `name` is required. All keys are then passed to the router link which enables using more complex routes.

### DIFF
--- a/changelog/unreleased/router-link-object
+++ b/changelog/unreleased/router-link-object
@@ -1,0 +1,7 @@
+CHANGE: Pass object as the target route
+
+We've changed the `targetRoute` prop in the `oc-resource` component to accept object instead of a string.
+This object is now accepting keys `name`, `params` and `query`. Only `name` is required.
+All keys are then passed to the router link which enables using more complex routes.
+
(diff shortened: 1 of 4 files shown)


changelog/unreleased/router-link-object
```diff
@@ -0,0 +1,7 @@
+CHANGE: Pass object as the target route
```
kulmann, line 1:
```suggestion
Change: Pass object as the target route
```

changelog/unreleased/router-link-object
```diff
@@ -0,0 +1,7 @@
...
+
+We've changed the `targetRoute` prop in the `oc-resource` component to accept object instead of a string.
```
kulmann, line 3:
```suggestion
We've changed the `targetRoute` prop in the `oc-resource` component to accept an object instead of a string.
```

kulmann (changes requested):
Tests are failing ;-)

kulmann (approved):
🚀

PR #1112, opened 2021-02-23
Add key to resource slot

LukasHirt:
To enforce re-render in case the resource has received a preview.


LukasHirt:
@kulmann Not sure if this is worth a changelog item? There's no user-facing change...

kulmann (approved):
No changelog needed IMO. LGTM 👍

PR #1113, opened 2021-02-23
Remove leftover variable from debugging

LukasHirt:


kulmann (approved):
You need to update snapshots ;-)

PR #1115, opened 2021-02-24
Text utility classes and truncating resource names in oc-table-files

kulmann:
This PR enables truncated resource names inside the oc-table-files component and enforces other columns not to wrap. See screenshot.

Screenshot 2021-02-24 at 08 09 10


LukasHirt (approved):
:rocket:

PR #1122, opened 2021-02-26
Remove overwritten ui-kit class from OcSidebar.vue

pascalwengerter:
The *class="uk-height-1-1"* in OcSidebar.vue translates to `height: 100;` and is outruled by `.oc-sidebar`'s `height: 100vh;`. It is also only used in the design system repo, not in web.

@LukasHirt do we need a CHANGELOG item for such minor fixes/documentation corrections, too?


pascalwengerter:
Forgot to rebase to most recent upstream/master, sry

kulmann (approved):

PR #1133, opened 2021-03-03
Quickwin: Improve image & user component documentation

pascalwengerter:
Went for more semantic HTML that also looks slightly better :)


LukasHirt (approved):
🚀

PR #1141, opened 2021-03-11
Implement labels and descriptions for text input fields

JammingBen:
* Remove placeholders
* Add required label to text inputs
* Add optional description message to text inputs
* Add `aria-describedby` which either references the description (if given), or an error/warning (if given)
* Add `aria-invalid` property

### DIFF
--- a/changelog/unreleased/enhancement-textinput-labels-and-description
+++ b/changelog/unreleased/enhancement-textinput-labels-and-description
@@ -1,0 +1,5 @@
+Enhancement: Implement labels and descriptions for text input fields
+
+This also implies all the required accessibility changes.
+
+https://github.com/owncloud/owncloud-design-system/pull/1141
--- a/src/components/OcAccordion.vue
+++ b/src/components/OcAccordion.vue
@@ -204,7 +204,6 @@ To see documentation on how to use this component, see [oc-accordion-item](/#/oC
     </p>
   </oc-accordion-item>
   <oc-accordion-item title="Something else with content" description="And a subtitle" icon="file">
-      <p>Enter some text!</p>
       <oc-text-input label="Text"></oc-text-input>
   </oc-accordion-item>
 </oc-accordion>
--- a/src/components/OcTextInput.vue
+++ b/src/components/OcTextInput.vue
@@ -1,11 +1,10 @@
 <template>
   <div>
-    <label :id="labelId" class="oc-label" :for="inputId" v-text="label" />
+    <label class="oc-label" :for="inputId" v-text="label" />
     <input
       :id="inputId"
       v-bind="additionalAttributes"
       ref="input"
-      :aria-labelledby="labelId"
       :aria-invalid="ariaInvalid"
       :class="{
         'oc-text-input': !stopClassPropagation,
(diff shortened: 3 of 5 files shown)


fschade (approved):
LGMT 🚀

JammingBen:
@kulmann Applied the naming changes as discussed 🙂

kulmann (changes requested):
There are other components in the design system which rely on the `oc-text-input` component, like the search for example. Since those don't necessarily require a visible label, it makes sense IMO that we change the implementation of the components which depend on `oc-text-input` to not use it anymore + replicate the necessary styling (CSS duplication is fine here, since we want to move towards really standalone components). Could you look into that before we merge this PR here?

JammingBen:
> There are other components in the design system which rely on the oc-text-input component, like the search for example. Since those don't necessarily require a visible label, it makes sense IMO that we change the implementation of the components which depend on oc-text-input to not use it anymore + replicate the necessary styling (CSS duplication is fine here, since we want to move towards really standalone components). Could you look into that before we merge this PR here?

@kulmann Hmm but we would also need to implement the descriptions, errors etc. again for each text input. Not necessarily for the search component, but for modal e.g. Maybe it is easier to provide an option to hide the label?

kulmann:
> > There are other components in the design system which rely on the oc-text-input component, like the search for example. Since those don't necessarily require a visible label, it makes sense IMO that we change the implementation of the components which depend on oc-text-input to not use it anymore + replicate the necessary styling (CSS duplication is fine here, since we want to move towards really standalone components). Could you look into that before we merge this PR here?
> 
> @kulmann Hmm but we would also need to implement the descriptions, errors etc. again for each text input. Not necessarily for the search component, but for modal e.g. Maybe it is easier to provide an option to hide the label?

`OcModal` can continue to use the `OcTextInput` like it is, because it requires to have the same a11y features like a standalone `OcTextInput`. 👍

kulmann:
Since https://github.com/owncloud/owncloud-design-system/pull/1151 is merged now (which removed a dependency on the oc-text-input component), could you please rebase this PR and then have a look through the documentation if all the occurrences of `oc-text-input` in examples are still rendering correctly? I will do the same after your rebase.

JammingBen:
> Since #1151 is merged now (which removed a dependency on the oc-text-input component), could you please rebase this PR and then have a look through the documentation if all the occurrences of oc-text-input in examples are still rendering correctly? I will do the same after your rebase.

Rebased 🙂  Checking now

kulmann (approved):
Nice job! 🥇

JammingBen:
Checked `OcModal` and `OcAccordion`, looking good. Although the input in `OcAccordion` kinda has two labels now :D

image

kulmann:
> Checked `OcModal` and `OcAccordion`, looking good. Although the input in `OcAccordion` kinda has two labels now :D
> 
> image

true. you could remove the `<p>Enter some text!</p>` then.

kulmann (approved):

JanAckermann:
Enter some Text could be the descriptionMessage, but merged, next time 👍

kulmann:
> Enter some Text could be the descriptionMessage, but merged, next time 👍

Create a PR ;-) happy to have it changed.

PR #1144, opened 2021-03-12
Add label and messages to textarea inputs

JammingBen:
Similar to https://github.com/owncloud/owncloud-design-system/pull/1141:

* Remove placeholders
* Add required label to textareas
* Add optional description message to textareas
* Add aria-describedby which either references the description (if given), or an error/warning (if given)
* Add aria-invalid property

Unfortunately there was no message line implemented yet, so I had to include this first.


JanAckermann (approved):

src/components/OcTextarea.vue
```diff
@@ -1,51 +1,148 @@
...
-    class="uk-textarea"
-    :value="value"
-    :placeholder="placeholder"
-    :aria-label="label"
-    @input="$_ocTextArea_onInput($event.target.value)"
-    @focus="$_ocTextArea_onFocus($event.target.value)"
-    @keydown="$_ocTextArea_onKeyDown($event)"
-  />
+  <div>
+    <label :id="labelId" class="oc-label" :for="textareaId" v-text="label" />
```
JanAckermann, line 3: do we still need the label id?

kulmann: Not needed anymore

src/components/OcTextarea.vue
```diff
@@ -1,51 +1,145 @@
...
   name: "OcTextarea",
   status: "review",
   release: "1.0.0",
   props: {
     /**
-     * Text value of the form textarea.
+     * The ID of the element.
      */
-    value: {
+    id: {
```
JanAckermann, line 56: We have already an id here, replace the default with unique-id and rename old textareaId :+1:

JanAckermann (changes requested):

JanAckermann (approved):

kulmann (changes requested):
Could you introduce the new styles as a style tag within the `OcTextarea.vue` file (i.e. as a real single file component) instead of in a new scss file? You can do so by adding a `<style lang="scss"></style>` tag as third element (order is: `template`, `script`, `style`, `docs`).

kulmann (approved):

PR #1145, opened 2021-03-12
Added acessibility to checkbox component

JanAckermann:
Add **id** and **for** properties to label
Add **aria-checked**, **aria-labelledby,**  **aria-disabled** property to input

### DIFF
--- a/src/components/OcCheckbox.vue
+++ b/src/components/OcCheckbox.vue
@@ -1,17 +1,17 @@
 <template>
-  <label :for="id" :class="{ 'oc-cursor-pointer': !disabled }">
+  <span>
     <input
       :id="id"
-      v-model="$_ocCheckbox_model"
+      v-model="model"
       type="checkbox"
       name="checkbox"
-      :aria-label="label"
-      :class="$_ocCheckbox_classes"
+      :aria-checked="checked.toString()"
+      :class="classes"
       :value="option"
       :disabled="disabled"
     />
-    <span v-if="!hideLabel" :aria-hidden="true" v-text="label" />
-  </label>
+    <label :for="id" :class="labelClasses" v-text="label" />
+  </span>
 </template>
 <script>
 import { getSizeClass } from "../utils/sizeClasses"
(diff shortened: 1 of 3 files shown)


JammingBen:
Running `yarn test -u` should fix the tests

src/components/OcCheckbox.vue
```diff
@@ -2,15 +2,17 @@
...
-      :class="$_ocCheckbox_classes"
+      :aria-labelledby="labelId"
+      :aria-checked="checked.toString()"
+      :aria-disabled="disabled.toString()"
+      :class="classes"
       :value="option"
       :disabled="disabled"
     />
-    <span v-if="!hideLabel" :aria-hidden="true" v-text="label" />
+    <span v-if="!hideLabel" :id="labelId" :aria-hidden="true" v-text="label" />
```
kulmann, line 15: This is a bit more complicated. There are valid scenarios for both having a visual label and having no visual label. But in any scenario the checkbox needs to have an accessible label. I'd propose the following:
- get rid of the `hideLabel` prop
- keep the `label` prop (optional)
- add an `accessibleLabel` prop (required)
=> then have two spans: the one for `label` can be aria-hidden, the one for the `accessibleLabel` needs to be invisible for the user (use `oc-invisible-sr` class), but is the one actually used for the `aria-labelledby` prop on the checkbox. There will be situations then where the provided values for `label` and `accessibleLabel` are the same. I'm fine with that to not get an overcomplicated logic here.

JanAckermann: I don't like the idea:
then we have the top label and 2 spans inside, will come up with a better solution

JammingBen: Maybe I am missing something, but why not just remove `:aria-hidden="true"` from the current implementation? Then you can hide the label or not, but regardless it is used to label the checkbox (and therefore required ofc).

JanAckermann: have a look

src/components/OcCheckbox.vue
```diff
@@ -1,17 +1,24 @@
...
+      :class="classes"
       :value="option"
       :disabled="disabled"
     />
-    <span v-if="!hideLabel" :aria-hidden="true" v-text="label" />
-  </label>
+    <label
+      :id="labelId"
+      :for="id"
+      :class="{ 'oc-invisible-sr': hideLabel === true }"
```
kulmann, line 18:
```suggestion
      :class="{ 'oc-invisible-sr': hideLabel }"
```

changelog/unreleased/checkbox-acessibility
```diff
@@ -0,0 +1,5 @@
+Enhancement: Add aria properties to checkbox
```
kulmann, line 1: Please add the `enhancement-` prefix to the changelog filename.

src/components/OcCheckbox.vue
```diff
@@ -1,17 +1,17 @@
 <template>
-  <label :for="id" :class="{ 'oc-cursor-pointer': !disabled }">
+  <span :class="{ 'oc-cursor-pointer': !disabled }">
```
kulmann, line 2: As with the radiobutton, please move the `oc-cursor-pointer` class to the label and all the conditionally applied classes of the label into a computed property.

kulmann (changes requested):

src/components/OcCheckbox.vue
```diff
@@ -1,17 +1,17 @@
...
-      :aria-label="label"
-      :class="$_ocCheckbox_classes"
+      :aria-checked="checked.toString()"
+      :class="classes"
       :value="option"
       :disabled="disabled"
     />
-    <span v-if="!hideLabel" :aria-hidden="true" v-text="label" />
-  </label>
+    <label :for="id" :class="{ 'oc-invisible-sr': hideLabel }" v-text="label" />
```
kulmann, line 13: pointer class on the label got lost now.

kulmann (changes requested):

JanAckermann:
Fucked something up while rebasing :-1:

kulmann (approved):
eslint doesn't agree. ;-) but code LGTM 👍

PR #1147, opened 2021-03-15
Increase datepicker z-index

JammingBen:
Before fix:
image

Fixed:
image

### DIFF
--- a/src/styles/theme/oc-datepicker.scss
+++ b/src/styles/theme/oc-datepicker.scss
@@ -99,9 +99,10 @@
   &-popup {
     @extend .uk-position-fixed;
     @extend .uk-position-center;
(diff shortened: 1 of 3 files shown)


src/styles/theme/oc-datepicker.scss
```diff
@@ -99,10 +99,11 @@
   &-popup {
     @extend .uk-position-fixed;
     @extend .uk-position-center;
-    @extend .uk-position-z-index;
     @extend .uk-box-shadow-large;
     @extend .uk-background-default;
 
+    z-index: 9000;
```
kulmann, line 105: There is a `$z-index-modal` scss variable. Should be usable here. Could you try?

kulmann (approved):
👍

PR #1148, opened 2021-03-15
Add aria properties to radio input

JanAckermann:
Add **id** and **for** properties to label
Add **aria-checked**, **aria-labelledby,**  **aria-disabled** property to input

### DIFF
--- a/src/components/OcRadio.vue
+++ b/src/components/OcRadio.vue
@@ -1,17 +1,17 @@
 <template>
-  <label :for="id" :class="{ 'oc-cursor-pointer': !disabled }">
+  <span>
     <input
       :id="id"
-      v-model="$_ocRadio_model"
+      v-model="model"
       type="radio"
       name="radio"
-      :aria-label="label"
-      :class="$_ocRadio_classes"
+      :class="classes"
+      :aria-checked="(option === value).toString()"
       :value="option"
       :disabled="disabled"
     />
-    <span v-if="!hideLabel" :aria-hidden="true" v-text="label" />
-  </label>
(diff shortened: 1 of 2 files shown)


JanAckermann:
@kulmann, unfortunately, I found no possibility to set the checked property to the input element, I only can set it from outside the component,  set it inside the component, does not have any effect.

changelog/unreleased/radio-acessibility
```diff
@@ -0,0 +1,5 @@
+Enhancement: Add aria properties to radio input
```
kulmann, line 1: Please rename the changelog file to `enhancement-radio-accessibility` (add `enhancement-` prefix and fix typo in `accessibility`)

src/components/OcRadio.vue
```diff
@@ -1,17 +1,17 @@
 <template>
-  <label :for="id" :class="{ 'oc-cursor-pointer': !disabled }">
+  <span :class="{ 'oc-cursor-pointer': !disabled }">
```
kulmann, line 2: You need to move the cursor class to the label. Having the oc-cursor-pointer class on the span, I don't see a pointer on the label, only on the radiobutton itself. Moving it to the label again works. Not needed on the input, as it already has the pointer by default.

kulmann: IMO it makes sense to move the classes of the label into a computed property then, since there are zero to two classes for the label.

kulmann (changes requested):

kulmann (approved):
🚀

PR #1149, opened 2021-03-15
Implement label and description for the datepicker component

JammingBen:
* Remove placeholders
* Add required label to datepickers
* Add optional description message to datepickers
* Add aria-describedby which references the description
* Add aria-invalid property (which is basically always false as the datepicker does not allow invalid inputs -> no errors)

### DIFF
--- a/src/components/OcDatepicker.vue
+++ b/src/components/OcDatepicker.vue
@@ -1,9 +1,9 @@
 <template>
   <div>
-    <label class="oc-label" :for="datepickerId" v-text="label" />
+    <label class="oc-label" :for="id" v-text="label" />
     <datetime
       v-model="update"
-      :input-id="datepickerId"
+      :input-id="id"
       v-bind="additionalAttributes"
       class="oc-datepicker"
       :title="title"
@@ -66,7 +66,7 @@ export default {
     id: {
       type: String,
       required: false,
-      default: null,
+      default: uniqueId("oc-datepicker-"),
     },
     /**
      * The html element name used for the container of Example component.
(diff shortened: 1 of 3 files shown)


src/components/OcDatepicker.vue
```diff
@@ -50,20 +63,18 @@ export default {
    * type(s). See examples below:
    */
   props: {
+    id: {
```
JanAckermann, line 66: replace `datepickerId` with `id` :+1

src/components/OcDatepicker.vue
```diff
@@ -1,24 +1,37 @@
...
-    :type="type"
-    :phrases="phrases"
-    :min-datetime="minDatetime"
-    :max-datetime="maxDatetime"
-    :placeholder="placeholder"
-    @cancel="$_cancel()"
-    @input="$_input()"
-  ></datetime>
+  <div>
+    <label class="oc-label" :for="id" v-text="label" />
```
JanAckermann, line 3: I am not quite sure but we might need an value attribute for accessibility?
image

JammingBen: Need to do some research on that 👀 

But I don't think it's required. The value here is basically what the user inputs. The same goes for the `text-input-fields` component, where we also don't have a value attribute. The calendar-modal itself may be another story, though...

JanAckermann: Research done on my side -> not required

src/components/OcDatepicker.vue
```diff
@@ -107,18 +118,48 @@ export default {
...
+      return `${this.id}-message`
+    },
+    additionalAttributes() {
+      const additionalAttrs = {}
+      if (this.descriptionMessage) {
+        additionalAttrs["aria-describedby"] = this.messageId
+      }
+      return { ...this.$attrs, ...additionalAttrs }
+    },
+    ariaInvalid() {
```
kulmann, line 154: Not needed as of now. Please remove. :-) Same for the `aria-invalid` attribute in the template.

kulmann (changes requested):

kulmann (approved):
👍

PR #1151, opened 2021-03-15
Remove the OcTextInput component from the OcSearchBar component

JammingBen:
Also replace the loading spinner with the OcSpinner component in the course of this.

Related to https://github.com/owncloud/owncloud-design-system/pull/1141 as this PR removes placeholders from the `OcTextInput` component.

### DIFF
--- a/src/components/OcSearchBar.vue
+++ b/src/components/OcSearchBar.vue
@@ -50,6 +50,8 @@
  * ### Making sure a submit button exits
  *
  * Both a search and filter form does need a submit button, regardless if the button is visually perceivable or not. If a "buttonless" look is desired, use `button-hidden="true"`, which renders the button visually hidden.
+ *
+ * The `aria-label` of the loading spinner can be set via `customLoadingAccessibleLabel`. If not set, it will default to "Loading results".
  */
 export default {
   name: "OcSearchBar",
@@ -144,11 +146,12 @@ export default {
       default: false,
     },
     /**
-     * The aria label for the loading spinner
+     * A custom aria label for the loading spinner
      */
-    loadingAccessibleLabel: {
+    customLoadingAccessibleLabel: {
       type: String,
-      required: true,
+      required: false,
+      default: "",
     },
   },
   data: () => ({
@@ -172,6 +175,13 @@ export default {
       !this.buttonHidden && classes.push("oc-search-input-button")
 
       return classes
+    },
+    loadingAccessibleLabel() {
+      if (this.customLoadingAccessibleLabel) {
+        return this.customLoadingAccessibleLabel
+      }
+
+      return this.$gettext("Loading results")
     },
   },
   methods: {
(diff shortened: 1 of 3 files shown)


JammingBen:
I still need to investigate on "How-To-Aria" for the loading spinner.

kulmann:
Screenshot 2021-03-16 at 13 35 31

The border radius is visible when a button exists. That was not the case before.

src/components/OcSearchBar.vue
```diff
@@ -8,21 +8,15 @@
...
       <span v-if="icon" class="uk-form-icon">
         <oc-icon v-show="!loading" :name="icon" />
-        <div
-          v-show="loading"
-          :class="$_ocSpinner_class"
-          :aria-label="ariaLabel"
-          tabindex="-1"
-          role="img"
-        />
+        <oc-spinner v-show="loading" :size="spinnerSize" aria-label="Loading results" />
```
kulmann, line 11: I know that the aria-label was entirely broken before. Could you add a required property `loadingAccessibleLabel` (or similar naming) for this? Needs to be provided by the code that uses the searchbar.

JammingBen:
> The border radius is visible when a button exists. That was not the case before.

Fixed 👍 

Now the only thing that differs visually from the original implementation is the margin right to the small loading spinner. This happens due to the use of the `OcSpinner` component. I find it quite nice though, so I would just leave it that way?

kulmann (approved):
Nice 👍

src/components/OcSearchBar.vue
```diff
@@ -149,6 +143,13 @@ export default {
       required: false,
       default: false,
     },
+    /**
+     * The aria label for the loading spinner
+     */
+    loadingAccessibleLabel: {
+      type: String,
+      required: true,
```
kulmann, line 151: Thinking about it, this is a breaking change. We have `$gettext` available. Could you check if you can set a sensible default in a callback?

kulmann (changes requested):

src/components/OcSearchBar.vue
JammingBen: Please check again. `$gettext` cannot be used as default value (at least I didn't find a way to do so), thus I had to implement a computed property in addition.

kulmann: Ok, that's fine. But please use `loadingAccessibleLabel` for the prop and define another name for the computed property. The props are the public API of the component and should have good names. The names for the computed properties are not that important (of course still for us as developers of the design system ;-) ).

src/components/OcSearchBar.vue
```diff
@@ -158,20 +162,27 @@ export default {
...
+    inputClass() {
       const classes = ["oc-search-input"]
 
       this.icon && classes.push("oc-search-input-icon")
       !this.buttonHidden && classes.push("oc-search-input-button")
 
       return classes
     },
+    loadingAccessibleLabel() {
+      if (this.customLoadingAccessibleLabel) {
```
kulmann, line 180: Can be simplified to `this.customLoadingAccessibleLabel || this.$gettext("Loading results")`.

kulmann (approved):
Nice 🚀

PR #1152, opened 2021-03-17
Add aria label to the home icon within the breadcrumb

JammingBen:
Fixes one part of https://github.com/owncloud/web/issues/4329

> Starting point (home link) of bread crumb has no link text, neither visually nor programmatically.

### DIFF
--- a/src/components/OcBreadcrumb.vue
+++ b/src/components/OcBreadcrumb.vue
@@ -6,7 +6,7 @@
           v-if="home && index === 0"
           :to="item.to"
           class="uk-flex uk-float-left"
-          :aria-label="homeAccessibilityLabelValue"
+          :aria-label="homeAccessibleLabelValue"
         >
           <oc-icon name="home" />
         </router-link>
@@ -49,7 +49,7 @@ import OcIcon from "./OcIcon.vue"
  *
  * ## Accessibility
  *
- * You can provide an accessibility label to the home icon via `homeAccessibilityLabel`. If not set, it will default to "Go to Home".
+ * You can provide an accessibility label to the home icon via `homeAccessibleLabel`. If not set, it will default to "Go to Home".
  */
 export default {
   name: "OcBreadcrumb",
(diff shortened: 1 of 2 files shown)


JanAckermann (approved):

src/components/OcBreadcrumb.vue
```diff
@@ -69,6 +77,14 @@ export default {
       type: [Boolean],
       default: false,
     },
+    /**
+     * Aria-label for the home icon
+     */
+    homeAccessibilityLabel: {
```
kulmann, line 83:
```suggestion
    homeAccessibleLabel: {
```
Please rename - we have that naming (`accessibleLabel`) in other locations already and I would like to keep it consistent.

src/components/OcBreadcrumb.vue
```diff
@@ -81,16 +97,19 @@ export default {
...
       return [...this.items].reverse().slice(1)
     },
-    $_ocBreadcrumb_currentFolder() {
+    currentFolder() {
       if (this.items.length === 0 || !this.items) return false
 
       return [...this.items].reverse()[0]
     },
+    homeAccessibilityLabelValue() {
+      return this.homeAccessibilityLabel || this.$gettext("Go to Home")
```
kulmann, line 111: I don't know if it's too techie, but the breadcrumbs are used in others contexts as well, so `Go to root directory` would be more appropriate. But again, I have the feeling that that's too techie... what do you think?

kulmann (changes requested):

src/components/OcBreadcrumb.vue
JammingBen: I had the exact same thought. But decided against it because then we presume that home always links to the filelist. Otherwise "root directory" wouldn't make sense. If we can happily make that presumption, I find `Go to root directory` totally fine 🙂

kulmann: fine by me 👍

kulmann (approved):
👍

PR #1153, opened 2021-03-17
Release 3.2.1

kulmann:
This PR fixes a visual bug with displaying folder names that contain a dot (`.`). Most importantly it also adds a `resource-type` attribute to the oc-resource-name component which is required for testing. Otherwise we don't have a way to detect if a resource is a file or folder.

We need to create a release v3.2.1 from this PR and merge it back to master after the release. Do not merge until the release is done. Current master contains changes that we can't release, yet.

### DIFF
--- a/changelog/3.2.1_2021-03-17/bugfix-folder-names
+++ b/changelog/3.2.1_2021-03-17/bugfix-folder-names
@@ -1,0 +1,5 @@
+Bugfix: Folder names with dots
+
+Folder names with dots showed were separated into a basename and extension part in the oc-resource-name component, which has been fixed. Additionally, the oc-resource-name component now as a `resource-type` attribute, which reflects the `type` property of the displayed resource. This makes testing easier.
+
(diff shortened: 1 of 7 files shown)


JammingBen (approved):
LGTM 🚀

JammingBen:
...although it's quite confusing that the `name` and the filename in `path` differ. With this change, this is even more noticable. If there is no specific reason for that, I suggest to equalized those values.

image

kulmann:
> ...although it's quite confusing that the `name` and the filename in `path` differ. With this change, this is even more noticable. If there is no specific reason for that, I suggest to equalize those values.
> 
> image

Hit the `request changes` button, this is a bug, at least in the example ;-) Thanks for spotting it. Checking...

JammingBen (changes requested):
see https://github.com/owncloud/owncloud-design-system/pull/1153#issuecomment-801056362

JammingBen (approved):
Here we go!

kulmann:
Release done. Merging.

PR #1155, opened 2021-03-17
Fix nesting on html elements

JammingBen:
Fixes the nesting issues coming from the ODS mentioned in https://github.com/owncloud/web/issues/4345

For now this should be all in the ODS. However, it is possible some more adjustments are needed. Need to dig into Oc Web first 👀 

On a side note: Feels a bit weird to use flexbox on a `<span>` element in `OcAccordion`. But it seems to be legit. Keeping the `<div>` tags would require a change in the parent component, which currently can be any `<h1-6>` element.

### DIFF
--- a/changelog/unreleased/enhancement-element-nesting
+++ b/changelog/unreleased/enhancement-element-nesting
@@ -1,0 +1,5 @@
+Enhancement: Element nesting
+
+Fix the nesting for some html elements to make them w3c compliant.
+
+https://github.com/owncloud/owncloud-design-system/pull/1155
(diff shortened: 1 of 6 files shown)


pascalwengerter (approved):
LGTM!

JanAckermann (approved):

kulmann (approved):
👍

kulmann:
Could you rebase this PR? There were changes in the accordion item (probably the ones related to focus border).

JammingBen:
Done 🚀

PR #1161, opened 2021-03-19
Adjust icons and images to make them a11y compliant

JammingBen:
I implemented `Vue Inline SVG` in the course of this as it provides an easy way to add properties to SVGs. How a11y now works for icons and images:

> You can pass a label to the icon via the `accessibleLabel` property. The component will automatically set the 'alt' attribute to the value of the provided label (for img-tags) or add a title element which is also referenced by its ID via `aria-labelledby` (for svg-tags).
>
> Omit `accessibleLabel` if your icon has a decorative purpose only. In this case the component will: 1. set `aria-hidden` to true. 2. set `role` to presentation (for svg-tags). 3. set `focusable` to false (for svg-tags). 4. remove or empty all aria-related properties such as labels or alts.

Again removed some unnecessary prefixes from properties and methods.


JammingBen:
@kulmann Well... this one is way more work then I anticipated. I'm finished with the basics which now allow us to use icons and images in a a11y compliant way. The main changes for this were made in `OcIcon`. I also adjusted the docs in there to explain how to use it.

One key change to notice here: When omitting the `accessibleLabel` on an icon or image, it would now be interpreted as non-existent for screen readers. Which makes sense IMO, but that also means that we have to provide those labels on each icon/image. At the moment, this falls back to the label `icon`... which is pretty meaningless, but at least gets seen by screen readers.

What do you think? Keep that fallback? Or implement meaningful labels?

kulmann:
As far as I understand we can just hide icons and images for screen readers if they only serve cosmetic purposes. I would not expect to have an accessible label for an icon and some textual representation.

I.e. if there is an accessible label => use it for the svg directly. I there is none => aria hide.

Maybe @marcus-herrmann could shed some light on this?

marcus-herrmann:
What I see in the code, for example...

```
   <inline-svg
      :src="svgPath"
      :transform-source="transformSvgElement"
      :aria-hidden="accessibleLabel === ''"
      :aria-labelledby="accessibleLabel === '' ? null : svgTitleId"
      :focusable="accessibleLabel === '' ? 'false' : null"
      :role="accessibleLabel === '' ? 'presentation' : 'img'"
    ></inline-svg>
```

..looks resonable to me. The existence of accessibleLabel is the main flag which either decides a SVGs presentational or relevant-for-the-content role.

Maybe I would make aria-label a little more robust by having explicit boolean value strings:

`:aria-hidden="accessibleLabel === '' ? 'true' : 'false' "`

Regarding the image, lines 27 to 30:
* No need for aria-hidden here. A null alt (`alt=""`) hides the image already from screen readers
* No need for an ARIA role, it has this role already (``)
* No need for focusable=false. Even IE11 does not try to focus good old images. It tries to do so for SVGs, though. That's why it's needed there

src/components/OcIcon.vue
```diff
@@ -1,45 +1,65 @@
...
+      :focusable="accessibleLabel === '' ? 'false' : null"
+      :role="accessibleLabel === '' ? 'presentation' : 'img'"
+    ></inline-svg>
+  </component>
   <!-- eslint-enable vue/no-v-html -->
   <img
     v-else
     :src="iconUrl"
-    :aria-label="ariaLabel"
+    :alt="accessibleLabel === '' ? '' : accessibleLabel"
```
kulmann, line 27:
```suggestion
    :alt="accessibleLabel"
```

src/components/avatars/OcAvatar.vue
```diff
@@ -4,8 +4,10 @@
     :username="userName"
     :size="width"
     :src="src"
-    :alt="accessibleLabel"
-    :aria-hidden="accessibleLabel === ''"
+    :aria-label="accessibleLabel === '' ? null : accessibleLabel"
+    :aria-hidden="accessibleLabel === '' ? 'true' : 'false'"
```
kulmann, line 8: Can't we just fall back to `null` instead of `'false'` here as well, like with the other attributes?

src/components/resource/OcResource.vue
```diff
@@ -5,18 +5,10 @@
...
     />
-    <oc-icon
-      v-else
-      key="resource-icon"
-      :name="resource.icon"
-      variation="system"
-      size="large"
-      aria-hidden="true"
-    />
+    <oc-icon v-else key="resource-icon" :name="resource.icon" variation="system" size="large" />
```
kulmann, line 11: Needs a rebase now. The `system` variation doesn't exist anymore. Use `passive` instead, which is also the new default (so could could also just omit the variation attribute). See https://github.com/owncloud/owncloud-design-system/blob/917a6c8c54191c80c35d6aabb9314b03b61218f8/src/components/OcIcon.vue#L94

kulmann (changes requested):
Cool! Needs a rebase and some small changes. But still, awesome work!

JammingBen:
Alright, rebased successfully 👍

changelog/unreleased/enhancement-icon-and-image-accessibility
```diff
@@ -0,0 +1,3 @@
+Enhancement: Add properties to make icons and images a11y-compliant
```
pascalwengerter, line 1: Is this still an enhancement or possibly a breaking change with all the new stuff around icons/images? We'll release a new major version around the theming anyways so just asking for clarification

pascalwengerter (commented):
Looks awesome overall, just two quick questions from my side

src/components/OcAppSideBar.vue
```diff
@@ -70,9 +70,6 @@ export default {
 ```vue
   <template>
     <oc-app-side-bar>
-        <template slot="action">
-          <oc-button class="uk-float-right oc-mr-s"><oc-icon name="close" /></oc-button>
```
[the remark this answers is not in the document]
kulmann: it's just a documentation example. Questions remains the same though 😁

changelog/unreleased/enhancement-icon-and-image-accessibility
kulmann: good catch, it's a breaking change 👍

kulmann: or let's put it like this: it's a breaking change regarding accessibility, not functionality.

src/components/OcAppSideBar.vue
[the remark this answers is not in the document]
JammingBen: As far as I understand, these few lines just "re-implement" the default action, which is the close button. I stumbled upon this because I tested something on the close-button defined in `<template>`, but it did not render. I can't see any difference on small screens either way 🤔

changelog/unreleased/enhancement-icon-and-image-accessibility
JammingBen: You're totally right, will change this 🙂

src/components/OcAppSideBar.vue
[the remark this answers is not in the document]
pascalwengerter: Ah then it's nice to have them deleted actually!

kulmann (approved):
💪

PR #1165, opened 2021-03-22
Fix avatar group tooltip which only has links

kulmann:
This PR fixes the tooltip of the oc-avatar-group in case it was only displaying links.

We'll need to create a release from this branch and merge it back to master after the release. Do not merge until then.


LukasHirt (approved):
:rocket:

PR #1164, opened 2021-03-22
Make searchbar component a11y-compliant

JammingBen:
fixes https://github.com/owncloud/owncloud-design-system/issues/1160

PR #1166, opened 2021-03-22
Make alert component a11y-compliant

JammingBen:
Similar to the changes in https://github.com/owncloud/owncloud-design-system/pull/1164

* Make "close"-button of an alert accessible via keyboard
* Remove the "uk-close" icon as it is not a11y-compliant


JanAckermann (approved):

kulmann (approved):
Nice! 🚀

PR #1140, opened 2021-03-10
Update color scheme & initialize CSS variables in theme

pascalwengerter:
**Update:** Successfully replaced all SASS variables with custom CSS properties and the design system _mostly_ works as expected again. We'll have to double-check all components to reveal hidden `uikit` styles (and integrate those).

PR #1168, opened 2021-03-23
Create custom CSS props from themeOptions

pascalwengerter:
Basic building block of initializing the ODS with themeOptions, by creating custom CSS props and overwriting the default color values.

Testing for the positive already works, checking for (un-over-written) default values doesn't work yet.

Depends on
- #1140 
- [#4822 in web](https://github.com/owncloud/web/pull/4822)

### DIFF
--- a/src/system.js
+++ b/src/system.js
@@ -19,7 +19,7 @@ const System = {
   install(Vue, options = {}) {
     const themeOptions = options.tokens
 
-    for (const colorVar in themeOptions.colorPalette) {
+    for (const colorVar in themeOptions?.colorPalette) {
       document.querySelector(':root').style.setProperty("--oc-" + colorVar, themeOptions.colorPalette[colorVar])
     }
     components.forEach(component => Vue.component(component.name, component))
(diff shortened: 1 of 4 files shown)


src/system.js
```diff
@@ -6,7 +6,7 @@
  */
 
 // Define contexts to require
-const contexts = [require.context("@/components/", true, /\.vue$/)]
+const contexts = [require.context("./components/", true, /\.vue$/)]
```
pascalwengerter, line 9: @LukasHirt couldn't make this one work properly (guess it's a Webpack thing), maybe you can shed some lights on it?

LukasHirt: `@` is an alias created in webpack which is equal to the `src` path.

https://github.com/owncloud/owncloud-design-system/blob/917a6c8c54191c80c35d6aabb9314b03b61218f8/build/webpack.base.conf.js#L30

pascalwengerter: Since this doesn't work as expected and we're about to switch from Webpack to Rollup we've decided not to spend further effort on resolving it for now

kulmann (approved):

PR #1169, opened 2021-03-24
Fix progress-bar background-color variations

pascalwengerter:


kulmann (approved):

kulmann:
@pascalwengerter could you add the url of this PR to the changelog where you introduced the css vars?

PR #1173, opened 2021-03-25
Drop placeholders in favor of descriptionMessages in OcModal

pascalwengerter:
Making the behavior consistent with other usages of input-fields within the ODS


kulmann (approved):
👍

PR #1174, opened 2021-03-26
Add description message to autocomplete component

JammingBen:


changelog/unreleased/enhancement-autocomplete-description
```diff
@@ -0,0 +1,5 @@
+Enhancement: Add description message to autocomplete component
+
+The description message will basically replace the informative text which was displayed via placeholder. It is still possible to use a placeholder though.
```
pascalwengerter, line 3: "Still possible to use a placeholder" but prop got deleted?

JammingBen: Yes because of `v-bind="additionalAttributes"`. It basically returns additional attributes + attributes which are being passed to the element. Hence the placeholder property does not need to be defined. I guess it's a classical question about "Do we keep it implicit oder explicit?" :D

pascalwengerter (approved):
Awesome

JanAckermann:
Maybe add a error message as well ? :heart:

JanAckermann (changes requested):
Add error/warning message

JammingBen:
> Maybe add an error message as well ? ❤️

Good point! Depends on when @pascalwengerter is planning to release the next minor 🙂  I'd be happy to implement it, would probably take an additional hour or so. If that conflicts with the release, I'll take care afterwards.

pascalwengerter:
> > Maybe add an error message as well ? ❤️
> 
> Good point! Depends on when @pascalwengerter is planning to release the next minor 🙂 I'd be happy to implement it, would probably take an additional hour or so. If that conflicts with the release, I'll take care afterwards.

Go for it, better to have it now (and release can wait another hour)!

changelog/unreleased/enhancement-autocomplete-description
kulmann: Maybe change the wording from `It is still possible to use a placeholder though.` to `While it's still possible to use a placeholder if you need it, we encourage you to not use it anymore.`. A label is always required, a description message can be added if more context is needed.

kulmann (changes requested):

pascalwengerter (approved):
LGTM (pushed some small changes myself)

JanAckermann (approved):

kulmann (approved):
👍

PR #1176, opened 2021-03-26
Match datepicker input style with other existing inputs

JammingBen:
Another very quick win as I don't know when the "new datepicker" will be finished.


kulmann (approved):
🚀

PR #1183, opened 2021-03-30
[WIP] Accessible datepicker component

JammingBen:
@kulmann I don't know wether/when I can continue on this, but here is the current progress on the new datepicker. It actually works pretty good and has all the functionality the previous one had (at least I'm aware of). Still, some improvements could be made on the keyboard navigation though.

image


kulmann:
Kind of superseded by https://github.com/owncloud/owncloud-design-system/pull/1661 - although this PR here does a little bit more. Linked this PR in a ticket for future reference. Thanks for the good work! ❤️

PR #1184, opened 2021-03-30
Fix links count in avatar group

kulmann:
The oc-avatar-group was showing all link, ignoring the `maxDisplayed` property. Fixed by properly cutting off the items used for rendering avatars. Added some tests to verify the behaviour at least on the tooltips.


pascalwengerter (approved):
Works like a charms!

PR #1198, opened 2021-04-07
Add vue select

LukasHirt:
Screenshot 2021-04-07 at 16 19 48

I am on purpose not implementing any unit tests as the lib has its own unit tests and we do not do any custom magic on top of that. Even our test coverage is 100 for that file...

### DIFF
--- a/src/components/OcSelect.vue
+++ b/src/components/OcSelect.vue
@@ -51,8 +51,12 @@ export default {
       fill: var(--oc-input-color);
     }
 
-    &__dropdown-option--highlight {
-      background: var(--oc-input-color);
+    &__dropdown-option {
+      white-space: normal;
+
+      &--highlight {
+        background: var(--oc-input-color);
+      }
     }
 
     &__actions svg {
@@ -61,6 +65,10 @@ export default {
 
     &__clear svg {
       max-width: 10px;
+    }
+
+    &__open-indicator {
+      margin-top: 2px;
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/components/OcSelect.vue
```diff
@@ -0,0 +1,182 @@
...
+]
+
+export default {
+  data: () => ({
+    selected: options[0],
+    options
+  })
+}
+</script>
+<style scoped>
```
pascalwengerter, line 176: Scoped styles need to be avoided in order not to blow up the bundle size AFAIK?

pascalwengerter: Nevermind, this is an example

src/components/OcSelect.vue
```diff
@@ -0,0 +1,182 @@
...
+export default {
+  data: () => ({
+    selected: 'Apple'
+  })
+}
+</script>
+```
+
+### Prevent user from clearing the selection
+If we want to disable the clear button from the `oc-select`, we can set prop `clearbable` to false. This will also prevent clearing the selected value by hitting `delete`.
```
pascalwengerter, line 95: `clearbable` => `clearable`

pascalwengerter (changes requested):
Small typo, also wondering (and experimenting myself) if we can add sth like https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap to avoid having a horizontally scrollable dropdown?

pascalwengerter (approved):
LGTM 🎈

fschade (approved):
nice, LGMT 👍

PR #1199, opened 2021-04-07
use html-loader for svg images

fschade:
the pr changed the way how svg images get loaded in oc-icon.
Before this, svg images where transformed to data url's by webpack.
Now we use require to inline them which prevents the browser to do a XMLHttpRequest and no additional tweaking of content-security-police is required anymore.


pascalwengerter (approved):
LGTM 🎈

PR #1202, opened 2021-04-08
cleanup package.json and move all dependencies to peerDependencies

fschade:
before this pr the package.json contained a lot of unnecessary dependencies which where hard linked.
This pr moves them to the peerDependencies and let the consuming application decide which patch or minor version to use.

We also had a lot of unused dependencies before.

Because of the way dependency management works this pr introduces breaking changes and a major release is required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "design",
     "system"
   ],
-  "homepage": "https://owncloud.github.io/owncloud-design-system/",
+  "homepage": "https://owncloud.design/",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/owncloud/owncloud-design-system.git"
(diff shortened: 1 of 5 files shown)


LukasHirt (approved):
LGTM 👍

JammingBen (commented):
LGTM, although pipeline disagrees :D Run `yarn test -u` and see what's changed?

package.json
```diff
@@ -2,13 +2,25 @@
...
   "description": "ownCloud Design System is based on VueDesign Systems and is used to design ownCloud UI components",
+  "keywords": [
+    "vue design system",
+    "design system",
+    "vue.js",
+    "vue",
+    "design",
+    "system"
+  ],
+  "homepage": "https://owncloud.github.io/owncloud-design-system/",
```
pascalwengerter, line 13: Could be owncloud.design for a nice URL :)

pascalwengerter (changes requested):
Missing a changelog item and one suggestion to improve the homepage URL, otherwise seems good to go!

package.json
fschade: done, thanks.

pascalwengerter (approved):

PR #1207, opened 2021-04-09
Remove fixed width for login-logo

pascalwengerter:
Fixes cases like this

Screenshot_2021-04-09 accessDenied - ownCloud

### DIFF
--- a/changelog/unreleased/bugfix-fixed-logo-width
+++ b/changelog/unreleased/bugfix-fixed-logo-width
@@ -1,0 +1,8 @@
+Bugfix: Remove fixed login logo width
+
+The logo displayed on login/error pages had a fixed width and height, wich was fine for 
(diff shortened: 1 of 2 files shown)


kulmann (commented):
What if a logo is much taller than wide? As an alternative we could set a `max-height` and `max-width` of 200px each. That would still keep the correct aspect ratio while not shrinking to tiny sizes.

pascalwengerter:
> What if a logo is much taller than wide? As an alternative we could set a `max-height` and `max-width` of 200px each. That would still keep the correct aspect ratio while not shrinking to tiny sizes.

Wasn't aware using the combination of max-width & max-height keeps the aspect ratio in a proper way, nice! Implemented and updated the PR accordingly.

kulmann (approved):
👍

PR #1210, opened 2021-04-13
Change uk-muted => oc-muted color class

pascalwengerter:


src/styles/theme/oc-text.scss
```diff
@@ -18,7 +18,7 @@ a {
 }
 
 .uk-text-meta,
-.uk-text-muted {
+.oc-text-muted {
```
kulmann, line 21: could you either have `*-text-meta` and `*-text-muted` a) with both prefix (`oc` for us and `uk` for overwriting usages of the uikit version) or b) with only the `oc` prefix and remove all the `uk-` usages in web in a subsequent PR? :-)

pascalwengerter: Highly in favor of dropping the `*-text-meta` one since it's duplicate and less verboose and settling for `oc-text-muted` in ODS & web (WIP PR here: https://github.com/owncloud/web/pull/4965, will update both accordingly)

kulmann: :heart:

kulmann (approved):

PR #1219, opened 2021-04-15
OCTable and OCTableFiles sorting

fschade:
This PR brings back sorting to the OCTable and OCTableFiles, it also fixes some tests which errored in the before.

### DIFF
--- a/changelog/unreleased/enhancement-table-sorting
+++ b/changelog/unreleased/enhancement-table-sorting
@@ -1,0 +1,6 @@
+Enhancement: Table sorting
+
+The latest version of OcTable and OcTableFiles had no more sorting.
+We brought it back. It can be enabled by simply adding a `sortable: true` to table fields.
+
+https://github.com/owncloud/owncloud-design-system/pull/1219
--- a/src/components/table/OcTable.vue
+++ b/src/components/table/OcTable.vue
@@ -1,12 +1,13 @@
 <template>
-  <table :class="tableClasses">
+  <table v-bind="extractTableProps()">
     <oc-thead v-if="hasHeader">
       <oc-tr>
         <oc-th
-          v-for="field in fields"
+          v-for="(field, index) in fields"
           :key="`oc-thead-${field.name}`"
-          v-bind="extractThProps(field)"
-          :style="{ top: sticky ? headerPosition + 'px' : null }"
+          v-bind="extractThProps(field, index)"
+          @click.native="$emit(constants.EVENT_THEAD_CLICKED, field)"
+          @keydown.enter.native="$emit(constants.EVENT_THEAD_CLICKED, field)"
         >
           <slot v-if="field.headerType === 'slot'" :name="field.name + 'Header'" />
           <template v-else>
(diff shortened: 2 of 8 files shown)


changelog/unreleased/enhancement-table-sorting
```diff
@@ -0,0 +1,6 @@
+Enhancement: Table sorting
+
+The latest version of OcTable and OcTableFiles had no more sorting.
+With this pr we bring it back by simply adding a sortable true to the table fields.
+
+https://github.com/owncloud/owncloud-design-system/pull/1216
```
kulmann, line 6:
```suggestion
https://github.com/owncloud/owncloud-design-system/pull/1219
```

changelog/unreleased/enhancement-table-sorting
```diff
@@ -0,0 +1,6 @@
+Enhancement: Table sorting
+
+The latest version of OcTable and OcTableFiles had no more sorting.
+With this pr we bring it back by simply adding a sortable true to the table fields.
```
kulmann, line 4:
```suggestion
We brought it back. It can be enabled by simply adding a `sortable: true` to table fields.
```
Changelog text should be a self contained short summary of a PR, not mentioning `this pr`. Otherwise it suggests that you need to look at the PR to understand the changelog.

src/components/table/OcTable.vue
```diff
@@ -395,40 +430,40 @@ export default {
...
           type: "callback",
           callback: function(timestamp) {
-            const date = new Date(timestamp * 1000)
-            const hours = date.getHours()
-            const minutes = "0" + date.getMinutes()
-            const seconds = "0" + date.getSeconds()
-            return hours + ":" + minutes.substr(-2) + ":" + seconds.substr(-2)
-          },
-        }]
+            const date = new Date(timestamp * 1000);
```
kulmann, line 439: We don't use semicolons in ODS 🤔

src/components/table/OcTable.vue
```diff
@@ -164,6 +172,9 @@ export default {
     },
   },
   computed: {
+    tableData() {
+      return this.sortedData || this.data
```
kulmann, line 176: Looks to me like `this.sortedData` always returns, well, data. Why do we need the `|| this.data`? 🤔

src/components/table/mixins/sort.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+  },
+  created() {
+    if (this.isSortable) {
+      this.$on(EVENT_THEAD_CLICKED, this.handleSort)
+    }
+  },
+  computed: {
+    sortedData() {
+      const data = [...this.data]
```
kulmann, line 21: You could have an early return `this.data` if `this.sortBy === null`

src/components/table/mixins/sort.js
```diff
@@ -0,0 +1,62 @@
...
+      const { sortable } = field
+      return !!sortable
+    },
+    handleSort(field) {
+      if (!this.fieldIsSortable(field)) {
+        return
+      }
+
+      if (field.name === this.sortBy) {
+        this.sortDir =
```
kulmann, line 44: Add a return after assigning `this.sortDir`. The `this.sortBy` already is `field.name` anyway, so I consider this a good use case for early returns.

src/components/table/mixins/sort.js
```diff
@@ -0,0 +1,62 @@
...
+          this.sortDir === SORT_DIRECTION_ASC ? SORT_DIRECTION_DESC : SORT_DIRECTION_ASC
+      }
+      this.sortBy = field.name
```
kulmann, line 47: Please reset `this.sortDir` to our default value (whatever that is) in the next line. When I change the sort field, I'd expect the sort direction to reset itself. I.e. something like `this.sortDir = SORT_DIRECTION_ASC`.

src/components/table/helpers/mixin-sort.js
```diff
@@ -0,0 +1,60 @@
+import { EVENT_THEAD_CLICKED } from "./constants"
```
kulmann, line 1: Delete this file. ;-) It's the old namespaced one in the helpers folder. The correct one is the one in the mixins folder, was we discussed earlier today.

kulmann (changes requested):
Excited to see this merged 🤩 Some small changes needed...

src/components/table/OcTable.vue
fschade: ugh, thanks

src/components/table/OcTable.vue
fschade: it's a fallback chain because if we have the custom sorters as discussed yesterday and one mixinData returns nothing it falls back to the default data.

```
this.sortedData || this.paginatedData || this.data

File: src/components/table/mixins/sort.js (hunk 5 above)
Comment (fschade) [reply src/components/table/mixins/sort.js:21]: fixed, good catch. now tableData makes more sense also for the moment

File: src/components/table/helpers/mixin-sort.js (hunk 8 above)
Comment (fschade) [reply src/components/table/helpers/mixin-sort.js:1]: gone, thanks

File: src/components/table/mixins/sort.js (hunk 7 above)
Comment (fschade) [reply src/components/table/mixins/sort.js:47]: if the second column gets sorted the data from column 1 depends on it and is not sorted anymore. this is why all other columns expect the sorted one gets 

```
aria-sort="none"
```

https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaSort

src/components/table/OcTable.vue
kulmann: sorted data will be paginated as well... the fallback feels weird to me

fschade: please have a look at latest commit, i think it make now more sense

kulmann: 👍

kulmann (approved):

PR #1222, opened 2021-04-16
Use newer nodejs images in CI

pascalwengerter:
Publish of `v6.0.0` failed in CI with an error that I suspect to be due to an old node version in the CI pipeline

Error:
```sh
SyntaxError: Unexpected token '.'

in

color.lighten(parseInt(attributes?.transform?.lighten) || 0)
```

in https://drone.owncloud.com/owncloud/owncloud-design-system/2527/2/2

We used webhippie's latest node image which was updated 10 months ago anyways: https://hub.docker.com/r/webhippie/nodejs/tags?page=1&ordering=last_updated

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -23,7 +23,7 @@ def testing(ctx):
         'steps': [
             {
                 'name': 'dependencies',
-                'image': 'owncloudci/nodejs:15',
+                'image': 'owncloudci/nodejs:14',
                 'pull': 'always',
                 'commands': [
                     'yarn install'
@@ -31,7 +31,7 @@ def testing(ctx):
             },
             {
                 'name': 'eslint',
-                'image': 'owncloudci/nodejs:15',
+                'image': 'owncloudci/nodejs:14',
                 'pull': 'always',
                 'commands': [
                     'yarn lint:eslint',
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -114,7 +114,7 @@ def build(ctx):
         'steps': [
             {
                 'name': 'build-docs',
-                'image': 'webhippie/nodejs:latest',
+                'image': 'owncloudci/nodejs:12',
```
fschade, line 117: 'image': 'owncloudci/nodejs:15',

.drone.star
```diff
@@ -123,7 +123,7 @@ def build(ctx):
             },
             {
                 'name': 'build-system',
-                'image': 'webhippie/nodejs:latest',
+                'image': 'owncloudci/nodejs:12',
```
fschade, line 126: 'image': 'owncloudci/nodejs:15',

fschade (dismissed):
we already use 'image': 'owncloudci/nodejs:15', this should be unique.

LukasHirt:
> we already use 'image': 'owncloudci/nodejs:15'

Don't use v15 pls. It has EOL planned for June this year. Use even version instead - v14 is the current active LTS

pascalwengerter:
> > we already use 'image': 'owncloudci/nodejs:15'
> 
> Don't use v15 pls. It has EOL planned for June this year. Use even version instead - v14 is the current active LTS

Got it. `:latest` is also not an option then, right? ;D

kulmann (approved):

PR #1224, opened 2021-04-19
Add more props to custom CSS loading

pascalwengerter:


src/system.js
```diff
@@ -14,16 +14,23 @@ contexts.forEach(context => {
   context.keys().forEach(key => components.push(context(key).default))
 })
 
+function initializeCustomProps(tokens, prefix) {
+  for (const param in tokens) {
```
kulmann, line 18: Could you add an early return above the for loop, checking if tokens are provided?

src/system.js
```diff
@@ -14,16 +14,25 @@ contexts.forEach(context => {
   context.keys().forEach(key => components.push(context(key).default))
 })
 
+function initializeCustomProps(tokens, prefix) {
+  if (typeof tokens === "undefined" || Object.keys(tokens).length === 0) {
```
kulmann, line 18:
```suggestion
  if (!tokens) {
```
should be sufficient

pascalwengerter: Love it!

This only works for nonexistent tokens, if the token object is empty (`spacing: {}`) it doesn't return early but loops for...no content - whatever...

kulmann: yep, looping on empty input is fine :-)

kulmann (approved):
🚀

PR #1225, opened 2021-04-19
use yaml as token base

fschade:
- use yaml instead of json
- extract token transforms and formats
- use filters to differentiate between ods and docs


kulmann (approved):
👍

Issue #1170, opened 2021-03-24
Breadcrumb dropdown is not a11y-compliant


JammingBen:
The breadcrumb shrinks to a dropdown when the screen is <640px.

* Needs to be accessible via keyboard
* Needs a focus

Issue #866, opened 2020-09-16
Tooltips for breadcrumb items


PVince81:
- Make it possible to add a tooltip to the home icon
- When ellipsized, display the text as tooltip

PVince81:
Pre-requisite to make it easier: https://github.com/owncloud/owncloud-design-system/issues/874

Issue #849, opened 2020-08-27
Do not assume home icon in breadcrumbs is always link


LukasHirt:
When displaying the home icon instead of string for the first item in breadcrumbs we shouldn't assume that it will be always link. Sometimes it can be a button which means that we also need to support click event on that icon.

LukasHirt:
Related to #695

PR #1228, opened 2021-04-19
Enhance oc-breadcrumb a11y 

JanAckermann:
* Replace unordered list element with ordered list element
* Add nav surrounding tag
* Add aria-current='page' to the last element 
* Remove default home breadcrumb 
* Allow keyboard navigation (as well for mobile)

fixes https://github.com/owncloud/owncloud-design-system/issues/866
fixes https://github.com/owncloud/owncloud-design-system/issues/849
fixes https://github.com/owncloud/owncloud-design-system/issues/1170

### DIFF
--- a/changelog/unreleased/change-oc-breadcrumb-a11y
+++ b/changelog/unreleased/change-oc-breadcrumb-a11y
@@ -1,0 +1,8 @@
+Change: Improve accessibility of oc-breadcrumb component
+
+In order to enhance accessibility oc breadcrumb has been wrapped into a <nav> tag.
+The <ul> element changed to <ol> element, furthermore aria-current=page applies to the last item and
+the default home breadcrumb has been removed.
+Keyboard navigation has been fixed.
+
+https://github.com/owncloud/owncloud-design-system/pull/1228
--- a/src/components/OcBreadcrumb.vue
+++ b/src/components/OcBreadcrumb.vue
@@ -1,25 +1,32 @@
 <template>
-  <div :class="`oc-breadcrumb oc-breadcrumb-${variation}`">
-    <ul class="oc-breadcrumb-list">
+  <nav :class="`oc-breadcrumb oc-breadcrumb-${variation}`">
+    <ol class="oc-breadcrumb-list">
       <li v-for="(item, index) in items" :key="index" class="oc-breadcrumb-list-item">
-        <router-link
-          v-if="home && index === 0"
-          :to="item.to"
-          class="uk-flex uk-float-left"
-          :aria-label="homeAccessibleLabelValue"
+        <router-link v-if="item.to" :aria-current="getAriaCurrent(index)" :to="item.to">{{
+          item.text
+        }}</router-link>
+        <oc-button
+          v-else-if="item.onClick"
+          :aria-current="getAriaCurrent(index)"
+          appearance="raw"
+          @click="item.onClick"
         >
-          <oc-icon name="home" />
-        </router-link>
-        <router-link v-else-if="item.to" :to="item.to">{{ item.text }}</router-link>
-        <a v-else-if="item.onClick" @click="item.onClick">{{ item.text }}</a>
-        <span v-else v-text="item.text" />
+          {{ item.text }}
+        </oc-button>
+        <span v-else :aria-current="getAriaCurrent(index)" v-text="item.text" />
       </li>
-    </ul>
+    </ol>
     <div class="oc-breadcrumb-drop">
-      <label class="oc-breadcrumb-drop-label">
+      <label
+        ref="mobileDropdown"
+        tabindex="0"
+        class="oc-breadcrumb-drop-label"
+        @keydown.enter="$refs.mobileDropdown.click()"
+      >
         <span
           v-if="currentFolder"
           class="oc-breadcrumb-drop-label-text"
+          aria-current="page"
           v-text="currentFolder.text"
         />
         <oc-icon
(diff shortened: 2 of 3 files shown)


pascalwengerter (commented):
Love it! Is it really a breaking change, though?

The `home` in the example can be deleted, and could you check for related issues in the `web` and `ods` repos? I can vagually remember something around the home icon that maybe is solved by this!

JanAckermann:
@pascalwengerter removed the unnecessary stuff.
From my POV, it is breaking, if you had a default home before which is now missing

JanAckermann:
@kulmann merge ? :rocket:

changelog/unreleased/change-oc-breadcrumb-a11y
```diff
@@ -0,0 +1,8 @@
+Change: Advance accessibility to oc-breadcrumb component
```
kulmann, line 1: Could you rewrite the changelog item in a way that it reflects `what we did` instead of `what we will do` please? :-) Like `For enhanced accessibility of the oc-breadcrumb component we wrapped it in a <nav> tag.` etc.

src/components/OcBreadcrumb.vue
```diff
@@ -1,25 +1,31 @@
...
       <li v-for="(item, index) in items" :key="index" class="oc-breadcrumb-list-item">
-        <router-link
-          v-if="home && index === 0"
-          :to="item.to"
-          class="uk-flex uk-float-left"
-          :aria-label="homeAccessibleLabelValue"
+        <router-link v-if="item.to" tabindex="0" :to="item.to">{{ item.text }}</router-link>
+        <a
+          v-else-if="item.onClick"
+          tabindex="0"
```
kulmann, line 8: `tabindex="0"` is the default for an `a` tag. Should not be needed, as long as it has a non-empty `href`. For the `router-link` I'm not sure if it's needed (didn't check).

kulmann (changes requested):

src/components/OcBreadcrumb.vue
JanAckermann: Let me go again trough this, IHMO in this particular case the <a> with item.onClick never will have a href here

JanAckermann: According to the VUE docs, tabindex might not be necessary for router-link:
```
<router-link> is the component for enabling user navigation in a router-enabled app. The target location is specified with the to prop. It renders as an <a> tag with correct href by default
```

https://router.vuejs.org/api/#router-link

kulmann: Woah, didn't pay close enough attention. If it has a click handler it should not be an `a` tag. Could you check the usage in github.com/owncloud/web and github.com/owncloud/file-picker if a button would be semantically correct here?

JanAckermann: sounds suitable, wondering if the oc-button can also be aria-current=page?

kulmann: Should work. All attributes that you apply on a component will be forwarded to the root native component within the component template.

JanAckermann: @kulmann this needs a bit of css adjustments, should we dig into that?

changelog/unreleased/change-oc-breadcrumb-a11y
```diff
@@ -0,0 +1,8 @@
+Change: Extend accessibility to oc-breadcrumb component
+
+Due to enhance accessibility to the oc breadcrumb has been wrapped into a <nav> tag.
```
pascalwengerter, line 3: "In order to enhance [...] ~to~ the oc breadcrumb [...]" reads nicer :)

src/components/OcBreadcrumb.vue
```diff
@@ -29,19 +34,26 @@
...
             <router-link v-if="item.to" :to="item.to">{{ item.text }}</router-link>
-            <a v-else-if="item.onClick" @click="item.onClick">{{ item.text }}</a>
+            <oc-button
+              v-else-if="item.onClick"
+              justify-content="left"
+              appearance="raw"
+              @click="item.onClick"
+              >{{ item.text }}</oc-button
+            >
             <span v-else v-text="item.text" />
```
pascalwengerter, line 47: Would it make sense to add `:aria-current="getAriaCurrent(index)"` here, too?

changelog/unreleased/change-oc-breadcrumb-a11y
```diff
@@ -0,0 +1,8 @@
+Change: Extend accessibility to oc-breadcrumb component
```
pascalwengerter, line 1: "Improve acccessibility of" or "Extend accessibility of", also make sure the headline isn't too long (otherwise CI fails 🤦🏽 )

pascalwengerter (changes requested):
Small wording changes, almost good to go!

JanAckermann:
@pascalwengerter Thx for the hint, can you have a closer look to the css styles as we introduced oc-button instead of <a> some minor changes was necessary

pascalwengerter:
> @pascalwengerter Thx for the hint, can you have a closer look to the css styles as we introduced oc-button instead of some minor changes was necessary

Sure, let me check it locally

pascalwengerter:
> @pascalwengerter Thx for the hint, can you have a closer look to the css styles as we introduced oc-button instead of < a > some minor changes was necessary

@JanAckermann two things I noticed:
- There's no `:hover` styles for the `onClick` breadcrumb element. Was that a conscious decision? Easily fixed in line 140 and 179
- Can't navigate through the breadcrumbs using the keyboard (tab key), it only recognizes the button..

JanAckermann:
@pascalwengerter 
> Can't navigate through the breadcrumbs using the keyboard (tab key), it only recognizes the button..
Yup, expected, this is because the items.to doesn't provide a known path to the vue router, if it is a valid link, vue provides a href, only tags with hrefs are tab sensitive....

src/components/OcBreadcrumb.vue
```diff
@@ -158,14 +168,26 @@ export default {
...
+        font-size: 0.875rem;
       }
 
       > li a:hover,
       > li span:hover {
         color: var(--oc-color-swatch-brand-default);
       }
+
+      > li a:focus {
+        outline: -webkit-focus-ring-color auto 1px;
```
JanAckermann, line 183: @pascalwengerter I think we need to move this to oc-drop, right?

JanAckermann:
@pascalwengerter 
>There's no :hover styles for the onClick breadcrumb element. Was that a conscious decision? Easily fixed in line 140 and 179
span was also missing 140

JanAckermann:
@kulmann after changing the a element to a button element, 

the task in ods files-table comes to my mind:
- file-name: should be a link, not a button.

So now we have some inconsistent behaviour :cry:

src/components/OcBreadcrumb.vue
pascalwengerter: Could you try moving everything inside `&-drop {` there and see how hard it is to make it work? :D

kulmann:
> @kulmann after changing the a element to a button element,
> 
> the task in ods files-table comes to my mind:
> 
> * file-name: should be a link, not a button.
> 
> So now we have some inconsistent behaviour 😢

Clarified in a call that this was a misunderstanding. Whether it's a link or a button is always depending on context (if it's navigation or an action).

src/components/OcBreadcrumb.vue
JanAckermann: https://github.com/owncloud/owncloud-design-system/pull/1230 
done here

JanAckermann: @pascalwengerter if #1230 finally works and get merged I will remove this line here

JanAckermann: removed

JanAckermann:
@kulmann @pascalwengerter 
Ready to merge :rocket:

kulmann (approved):
LGTM 🚀

pascalwengerter (approved):
LGTM 🙌🏽

PR #1229, opened 2021-04-21
Enhance accessibility for the files table and its contents

JammingBen:
* Add accessible description in case a resource link opens in a new window
* Add accessible description for status indicators
* Adjust some of the labels to make them more descriptive
* Fix broken outline (focus) on directories

Note that the keyboard navigation for directories in ODS will most likely not work because the `router-link` component fails to transform `to=""` to `href=""`. It's working in Web though.

### DIFF
--- a/src/components/OcStatusIndicators.vue
+++ b/src/components/OcStatusIndicators.vue
@@ -7,6 +7,7 @@
         :key="indicator.id"
         class="oc-status-indicators-indicator"
         :aria-label="indicator.label"
+        :aria-describedby="getIndicatorDescriptionId(indicator)"
         :uk-tooltip="indicator.label"
         variation="passive"
         appearance="raw"
@@ -22,8 +23,16 @@
         class="oc-status-indicators-indicator"
         :name="indicator.icon"
         :accessible-label="indicator.label"
+        :aria-describedby="getIndicatorDescriptionId(indicator)"
         :uk-tooltip="indicator.label"
         variation="passive"
+      />
+      <span
+        v-if="getIndicatorDescriptionId(indicator)"
+        :id="getIndicatorDescriptionId(indicator)"
+        :key="indicator.id"
+        class="oc-invisible-sr"
+        v-text="indicator.accessibleDescription"
       />
     </template>
   </div>
@@ -32,6 +41,7 @@
 <script>
 import OcIcon from "./OcIcon.vue"
 import OcButton from "./OcButton.vue"
+import uniqueId from "../utils/uniqueId"
 
 /**
  * Status indicators which can be attatched to a resource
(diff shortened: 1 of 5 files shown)


src/components/OcStatusIndicators.vue
```diff
@@ -72,6 +82,17 @@ export default {
     hasHandler(indicator) {
       return Object.prototype.hasOwnProperty.call(indicator, "handler")
     },
+    getIndicatorDescriptionId(indicator) {
+      if (!indicator.accessibleDescription) {
+        return null
+      }
+
+      if (!indicator.accessibleDescriptionId) {
```
kulmann, line 90: What is the use case for providing an `accessibleDescriptionId` via data? Since it's always created within this component it would be fine by me to always create a random one, as long as it's unique.

src/components/table/OcTableFiles.vue
```diff
@@ -23,7 +23,7 @@
     <template #select="{ item }">
       <oc-checkbox
         :id="`oc-table-files-select-${item.id}`"
-        :label="$gettext('Select')"
+        :label="$gettext('Select resource')"
```
kulmann, line 26: The item has information available whether it's a folder or not. Could you go the extra mile an introduce a computed prop which has `$gettext('Select folder')` for a folder and `$gettext('Select file')` otherwise? Context: we had some usability tests recently and they unveiled that it's not directly clear what we mean by `resource`.

kulmann: Property of the item is `type` and has the value `folder` in case of a folder.

kulmann (changes requested):

src/components/OcStatusIndicators.vue
JammingBen: Maybe I get it wrong, but each `indicator` has its own `accessibleDescriptionId` which needs to be "bound" to it. Hence we cannot provide a "global" unique id here.

src/components/table/OcTableFiles.vue
JammingBen: Great idea :) I solved this via a method though, because we need to pass `resource` as parameter.

kulmann: True 🚀 nice!

src/components/OcStatusIndicators.vue
kulmann: What I meant is that I don't see a reason why we would provide an element id from external, when the html is always coming from within this component (the span with the `accessibleDescription` that you introduced in this PR).

JammingBen: `accessibleDescriptionId` is not supposed to be provided external (although it can, technically). The only reason we attach it to `indicator` is because we need to store the id somewhere. The if-condition does that initially for each `indicator`. The next time we call that method for the exact same `indicator`, it may already have an id. Then we just return it.

kulmann: Ah, now I get it. Thanks. In that case I'd prefer to cache the id in the data object of this component.

```
data() {
  return {
    accessibleDescriptionIds: {}
  }
}
```
And then instead of modifying `indicator.accessibleDescriptionId`, set `this.accessibleDescriptionIds[indicator.id] = uniqueId("oc-indicator-description-")` and reuse that in subsequent calls.

JammingBen: Oh that's nice, and should make it more clear also 👍  Thanks!

kulmann (approved):
nice! 🚀

PR #1230, opened 2021-04-21
Reorder oc-drop example data, add focus border to li a elements

JanAckermann:


pascalwengerter (commented):
Works great in Chrome, but not in Firefox (buttons are tab-abled and the nav expands on hitting enter, but then you can't "tab through the links") and Safari seems completely broken, at least when checking in the design system docs locally...

JanAckermann:
@pascalwengerter can you check again Chrome and FF works now

JammingBen:
>  and Safari seems completely broken, at least when checking in the design system docs locally...

FF is completely broken for me, too :( I haven't seen an outline throughout the entire ODS.

pascalwengerter:
> @pascalwengerter can you check again Chrome and FF works now

Just did, still doesn't work. Was it the goal of this PR to fix it, anyways?

JammingBen:
> Was it the goal of this PR to fix it, anyways?

Since it seems to be a general problem, I'd say it's not scope of this PR.

pascalwengerter (approved):
Looks good for now 🤘🏽 focus management / tab navigation are somewhat broken in Firefox/Safari but that's out of scope

kulmann (approved):

PR #1237, opened 2021-04-23
Add modal focus trap

LukasHirt:
We've added [Vue focus trap library](https://github.com/posva/focus-trap-vue) to trap the keyboard navigation inside of the modal.
To enable the focus trap, use prop `focusTrapActive`.


pascalwengerter (approved):
Works like a charm 🍰

kulmann (approved):
Nice! And it already has support for vue3 🤩  LGTM

PR #1240, opened 2021-04-26
feat: add paddingX prop

LukasHirt:
We've added new prop called `paddingX` which can be used to set the padding along x axis of the `oc-table` or `oc-table-files`.


kulmann (approved):
Awesome 🚀

PR #1241, opened 2021-04-26
A11y improvements

kulmann:
Small a11y changes regarding the oc-icon and oc-accordion components.


LukasHirt (approved):
Snapshots are probably failing... I'm starting to think if they're actually worth it 🤷  Otherwise great! 🚀

PR #1244, opened 2021-04-27
Aria-hide images without alt tags

kulmann:
When the `alt` property of the oc-image is empty we now set `aria-hidden="true"`.


fschade (approved):
🦖 nice, LGMT

PR #1245, opened 2021-04-28
Move href into extra computed properties

JanAckermann:
While setting the type property to 'router-link' the empty href property will overwrite the href target with null.
This means that the router link still works but the correct target is not set in the < a > element.

### DIFF
--- a/src/components/OcButton.vue
+++ b/src/components/OcButton.vue
@@ -1,13 +1,12 @@
 <template>
   <component
     :is="type"
-    :href="href"
-    :to="to"
+    :v-bind="properties"
     :type="submit"
     :aria-label="ariaLabel"
     :class="$_ocButton_buttonClass"
     :disabled="disabled"
-    @click="$_ocButton_onClick"
+    v-on="handlers"
   >
     <!-- @slot Content of the button -->
     <slot />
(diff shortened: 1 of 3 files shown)


src/components/OcButton.vue
```diff
@@ -160,6 +160,16 @@ export default {
 
       return classes
     },
+
+    properties() {
```
kulmann, line 164: nice! could you also move the `to` prop into this?

kulmann (changes requested):
Could you also move `to` into the `properties` computed prop and the click handler into a `v-on` in a similar manner? That way we would really only have the bindings that are relevant for the one (out of three) active component type.

src/components/OcButton.vue
```diff
@@ -1,13 +1,12 @@
 <template>
   <component
     :is="type"
-    :href="href"
-    :to="to"
+    :v-bind="properties"
```
kulmann, line 4: `v-` directives don't have the leading colon. Please change to `v-bind` (instead of `:v-bind`)

src/components/OcButton.vue
```diff
@@ -1,13 +1,12 @@
...
     :type="submit"
     :aria-label="ariaLabel"
     :class="$_ocButton_buttonClass"
     :disabled="disabled"
-    @click="$_ocButton_onClick"
+    :v-on:click="$_ocButton_onClick"
```
kulmann, line 9: Misunderstanding. ;-) What I meant was that you introduce a `handlers` computed prop  and have `v-on="handlers"` here.
For the computed prop:
```js
const handlers = {}
if (this.type === "button") {
  handlers.click = this.$_ocButton_onClick
}
return handlers
```

This way we only have the click handler set at all when the component type is button. Otherwise is doesn't even have a click handler. You could even remove the type check within the click handler then.

kulmann (approved):
LGTM 👍

PR #1247, opened 2021-04-28
Remove close button on notifications

JammingBen:
Notifications will now automatically close after a certain amount of time which can be defined via property. Also fixed some links in the docs in addition to that.

Small improvements in the course of this:

* fixed some links in the docs
* removed prefixes from properties

### DIFF
--- a/src/components/OCNotificationMessage.vue
+++ b/src/components/OCNotificationMessage.vue
@@ -90,8 +90,15 @@ export default {
     },
   },
   mounted() {
+    /**
+     * Notification will be destroyed if timeout is set
+     */
+    if (!this.timeout && this.dismissible) {
+      return
+    }
+
     setTimeout(() => {
-      this.$emit("close")
+      this.close()
     }, this.timeout * 1000)
   },
   methods: {
--- a/src/components/OCNotificationMessage.vue
+++ b/src/components/OCNotificationMessage.vue
@@ -13,17 +13,6 @@
         {{ message }}
       </div>
     </div>
-    <oc-button
-      v-if="dismissible"
-      name="close"
-      appearance="raw"
-      :variation="iconVariation"
-      :aria-label="$gettext('Close notification')"
-      class="uk-position-top-right oc-mt-s oc-mr-s"
-      @click="close"
-    >
-      <oc-icon name="close" :variation="iconVariation" />
-    </oc-button>
   </div>
 </template>
 <script>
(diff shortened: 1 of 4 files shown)


JammingBen:
@kulmann I know that we handle the timeout in Web currently. But with the absence of a close button I'd figured it would be best to implement this in the component directly. Also it's more transparent IMO. What do you think? I would remove all the timeout related stuff in Web then.

pascalwengerter (approved):
LGTM, perhaps needs a final go by @kulmann regarding the self-closing-ness though

JanAckermann:
Lets go with a dismissible property instead (defaults to false)

src/components/OCNotificationMessage.vue
```diff
@@ -68,17 +64,46 @@ export default {
...
+     * Number of seconds the message shows. It will disappear after this time. Setting the timeout to 0 while the
+     * notification is dismissible will bypass the timeout.
+     */
+    timeout: {
+      type: Number,
+      required: false,
+      default: 5,
+    },
+    /**
+     * Property weather a close button will be shown or not
```
kulmann, line 77:
```suggestion
     * Whether a close button will be shown or not
```

kulmann (changes requested):
Please don't introduce the `dismissible` property. Get rid of the button entirely, instead. We don't want to take care of focus management for the notification message right now. By offering the prop, a dev might think that it's (a11y wise) safe to use it with the button.

src/components/OCNotificationMessage.vue
JammingBen: "Solved" due to the removal of this property.

JammingBen:
Removed the property and rebased in the course of this.

pascalwengerter (approved):
LGTM 😎

src/components/OCNotificationMessage.vue
```diff
@@ -68,17 +53,33 @@ export default {
       required: false,
       default: null,
     },
+    /**
+     * Number of seconds the message shows. It will disappear after this time.
+     */
+    timeout: {
+      type: Number,
+      required: false,
+      default: 5,
```
kulmann, line 62: Last tiny bit: please add a validation that the timeout has to be greater than 0. Otherwise it would be possible to create notifications that won't disappear, ever.

kulmann (approved):
LGTM! One small remark. After that good to go 🥳

PR #1250, opened 2021-04-29
Add description for avatars

JammingBen:
The description property is mandatory for avatar groups because an avatar group is hidden for screen readers. For stand-alone avatars this property can be used optionally though.

This tackles the following 2 a11y-requirements:

* hide avatar-group for screen readers, just use text which describes the infos for screen readers
* aria-hidden=true on avatar of share owner. introduce a text element describing the share ownership which is visible for screen readers only

The descriptions now need to be added in Web.


src/components/avatars/OcAvatar.vue
```diff
@@ -1,14 +1,21 @@
...
-    class="oc-avatar"
-    :username="userName"
-    :size="width"
-    :src="src"
-    :aria-label="accessibleLabel === '' ? null : accessibleLabel"
-    :aria-hidden="accessibleLabel === '' ? 'true' : null"
-    :focusable="accessibleLabel === '' ? 'false' : null"
-    :role="accessibleLabel === '' ? null : 'img'"
-  />
+  <div class="oc-avatar-wrapper">
```
pascalwengerter, line 2: This is not valid in terms of a11y/HTML validation, since we're using the avatar component inside a button component (e.g. user menu)

pascalwengerter (changes requested):
Please check out my changes in https://github.com/owncloud/web/pull/5034/files#diff-7d3f2fc3e6ffeecd31a7d68ae32a8b532cbea2130f3d6b7a9bb69aa8133dc754

TL;DR: We've made the avatar component wrapper in `web` hidden="true" since it's only visual sugar. From my understanding, we don't need to touch the OcAvatar because of this, your changes to the OcAvatarGroup are of course still valid!

JammingBen:
@pascalwengerter Thanks for the hint! Was wondering because the stand-alone avatars in the filelist we not `aria-hidden`... then I realized they also use the `oc-avatar-group` component 🤦‍♂️

JammingBen:
I also adjusted the `<oc-avatar-group>` usage in our `OcFileTable` component. Should be ready to review now 🚀

kulmann:
@JammingBen could you rebase this PR please?

src/components/avatars/OcAvatarGroup.spec.js
```diff
@@ -41,21 +41,23 @@ describe("OcAvatarGroup", () => {
         users,
         maxDisplayed: 2,
         isTooltipDisplayed: true,
+        accessibleDescription: "List of users",
       },
     })
 
-    expect(wrapper.attributes()["uk-tooltip"]).toMatch("Bob, Marie +3")
+    expect(wrapper.find('.oc-avatar-group').attributes()["uk-tooltip"]).toMatch("Bob, Marie +3")
```
kulmann, line 48: We don't have the `uk-tooltip` anymore, needs to be updated

src/components/avatars/OcAvatarGroup.spec.js
```diff
@@ -41,21 +41,23 @@ describe("OcAvatarGroup", () => {
...
   it("hides tooltip if not enabled", () => {
     const wrapper = shallowMount(Group, {
       propsData: {
         users,
+        accessibleDescription: "List of users",
       },
     })
 
-    expect(wrapper.attributes()["uk-tooltip"]).toBeFalsy()
+    expect(wrapper.find('.oc-avatar-group').attributes()["uk-tooltip"]).toBeFalsy()
```
kulmann, line 60: Same here, we don't have the uk-tooltip anymore, needs to be updated

src/components/avatars/OcAvatarGroup.spec.js
```diff
@@ -64,10 +66,11 @@ describe("OcAvatarGroup", () => {
         users,
         maxDisplayed: 3,
         isTooltipDisplayed: true,
+        accessibleDescription: "List of users",
       }
     })
 
-    expect(wrapper.attributes()["uk-tooltip"]).toMatch("Bob, Marie, John Richards Emperor of long names +2")
+    expect(wrapper.find('.oc-avatar-group').attributes()["uk-tooltip"]).toMatch("Bob, Marie, John Richards Emperor of long names +2")
```
kulmann, line 73: Same here, we don't have the uk-tooltip anymore, needs to be updated

src/components/avatars/OcAvatarGroup.spec.js
```diff
@@ -76,10 +79,11 @@ describe("OcAvatarGroup", () => {
       propsData: {
         users,
         isTooltipDisplayed: true,
+        accessibleDescription: "List of users",
       }
     })
 
-    expect(wrapper.attributes()["uk-tooltip"]).toMatch("Bob, Marie, John Richards Emperor of long names, link 0, link 1")
+    expect(wrapper.find('.oc-avatar-group').attributes()["uk-tooltip"]).toMatch("Bob, Marie, John Richards Emperor of long names, link 0, link 1")
```
kulmann, line 86: Same here, we don't have the uk-tooltip anymore, needs to be updated

src/components/avatars/OcAvatarGroup.vue
```diff
@@ -1,28 +1,26 @@
...
-        :name="link.name"
-        :accessible-label="link.name"
-      />
-    </template>
-    <oc-avatar-count v-if="isOverlapping" :count="users.length - maxDisplayed" />
+  <div>
+    <div
+      class="oc-avatar-group"
+      :class="{ 'oc-avatar-group-stacked': stacked }"
+      :uk-tooltip="tooltip"
```
kulmann, line 6: During rebasing, please make sure that you don't accidentally add the `uk-tooltip` again. ;-) We have `v-oc-tooltip` now for tooltips.

kulmann (changes requested):

src/components/avatars/OcAvatarGroup.spec.js
fschade: ```expect(wrapper.attributes()["uk-tooltip"]).toMatch("Bob, Marie +3")```

all those tests should be fine after rebase, already in master

JammingBen:
@kulmann Rebased 👍  Waiting for the tests now, but at least locally they passed.

src/components/table/OcTableFiles.vue
```diff
@@ -388,6 +390,54 @@ export default {
...
+        return `${this.$gettext("This file is shared with the users")} ${userList}.`
+      } else {
+        return `${this.$gettext("This file is shared via the links")} ${linkList}.`
+      }
+    },
+
+    getOwnerAvatarDescription(resource) {
+      let description
+      if (resource.type === "folder") {
+        description = this.$gettext("This folder is owned by")
```
kulmann, line 434: Please use `$gettextInterpolate` wherever you inject/concatenate additional variables to a string.

kulmann: Reason is simple: in other languages the name might not be at the end of the string but somewhere in the middle, depending on the respective grammar of the language. Placing the variable at the correct location is a responsibility of the translation, not for the code.

src/components/table/OcTableFiles.vue
```diff
@@ -388,6 +390,54 @@ export default {
 
       return this.$gettext("Select file")
     },
+
+    getSharedWithAvatarDescription(resource) {
```
kulmann, line 394: I'd like to reduce complexity of this. It's sufficient to have `Shared with {user-count} recipients and via {link-count} links`. Don't go into details with files/folders and don't include the names. If it's 200 names I don't want to hear all of them. ;-)

There is still complexity left. We need to:
- a) have proper pluralization (1 recipient / n recipients, 1 link / n links)
- b) shares, links, or both being present

Please look at `$ngettext` and `$gettextInterpolate` for solving this.

pascalwengerter: Could you check my updated version below?

pascalwengerter: @kulmann do we need to add support for other shareTypes than user and link? Like

```js
const groupCount
const guestCount
const remoteCount
```

pascalwengerter: Line 398/399 in this current file, seems a bit to simplistic from my POV

JammingBen: If we/you implement additional distinctions here, please also keep the other computed properties in mind 🙂 
 https://github.com/owncloud/owncloud-design-system/blob/master/src/components/avatars/OcAvatarGroup.vue#L112

JammingBen (commented):
Since I can't approve my own PR: Nice work @pascalwengerter ! 🚀  LGTM

src/components/table/OcTableFiles.vue
kulmann: The share types currently are `link` and `everything else`. That is fine. We can't name it `user` because of that.

src/components/table/OcTableFiles.vue
```diff
@@ -388,6 +390,50 @@ export default {
...
+      const resourceType =
+        resource.type === "folder" ? this.$gettext("folder") : this.$gettext("file")
+
+      const userCount = resource.sharedWith.filter(u => !u.link).length
+      const linkCount = resource.sharedWith.filter(u => !!u.link).length
+
+      const baseText = this.$gettext("This %{ resourceType } is shared ")
+      const userText =
+        userCount > 0
+          ? this.$ngettext("with %{ userCount } user", "with %{ userCount } users", userCount)
```
kulmann, line 404: Cannot be named `user` / `users`, because it represents users, groups, federated and guest shares.

pascalwengerter: Are we talking variable naming or the "plain text" or both?

Alternative suggestion would be "There are X link shares and Y other shares for this file/folder"?

kulmann: we're talking about the plain text. ;-) `other shares` also doesn't work because that doesn't work standalone.

@pmaier1 @tbsbdr  opinions?

kulmann: We talked yesterday about `This file is shared with 23 recipients` also not being good naming.

pmaier1: `This file is shared with X people` (1 person) `and has Y links.` ?

kulmann: @pmaier1 that's exactly the issue. It's also group sharing. So `1 person` / `x people` would be just wrong.

Alternative: make a third split into groups. I.e. `links`, `people` and `groups`.

kulmann: People would include normal user shares, federated shares and guest shares then. Which are all people.

pmaier1: Ah, sorry. `This file has X invites and Y links` could work if you don't want to split. Would need to be used consistently in other areas as well.

pascalwengerter: Updated to `"This %{ resourceType } is shared via X invite(s) and Y link(s)"`, hope that gets the message across 😎

changelog/unreleased/change-avatar-group-description
```diff
@@ -0,0 +1,5 @@
+Change: Add description for avatar groups
```
kulmann, line 1: IMO not a change, but an Enhancement

src/components/table/OcTableFiles.vue
```diff
@@ -388,6 +390,46 @@ export default {
...
+
+    getSharedWithAvatarDescription(resource) {
+      const resourceType =
+        resource.type === "folder" ? this.$gettext("folder") : this.$gettext("file")
+
+      const shareCount = resource.sharedWith.filter(u => !u.link).length
+      const linkCount = resource.sharedWith.filter(u => !!u.link).length
+
+      const baseText = this.$gettext("This %{ resourceType } is shared via ")
+      const shareText =
```
kulmann, line 402: Argh. This doesn't work. Gettext extracts the strings. Concatenating the **results** (!) of gettext will provide correct results for English, but not for other languages that have different grammar...

kulmann: Meaning: All combinations / concatenations need to be full strings.

kulmann (changes requested):

changelog/unreleased/change-avatar-group-description
pascalwengerter: Fixed

src/components/table/OcTableFiles.vue
pascalwengerter: Updated the code, is this the pragmatic solution you were looking after? 🤔 

And does `$gettextInterpolate` complain about unused variables?

pascalwengerter: "Explanation":

```js
switch (true) {
  case shareCount == 1 && linkCount == 0:
    // one share no link
    description = this.$gettext("This %{ resourceType } is shared via one invite")
    break
  case shareCount == 0 && linkCount == 1:
    // no share one link
    description = this.$gettext("This %{ resourceType } is shared via one link")
    break
  case shareCount == 1 && linkCount == 1:
    // one share one link
    description = this.$gettext(
      "This %{ resourceType } is shared via one invite and one link"
    )
    break
  case shareCount == 0 && linkCount > 1:
    // no share many links
    description = this.$gettext("This %{ resourceType } is shared via %{ linkCount } link")
    break
  case shareCount == 1 && linkCount > 1:
    // one share many links
    description = this.$gettext(
      "This %{ resourceType } is shared via one invite and %{ linkCount } links"
    )
    break
  case shareCount > 1 && linkCount == 0:
    // many shares no links
    description = this.$gettext(
      "This %{ resourceType } is shared via %{ shareCount } invites"
    )
    break
  case shareCount > 1 && linkCount == 1:
    // many shares one link
    description = this.$gettext(
      "This %{ resourceType } is shared via %{ shareCount } invites and one link"
    )
    break
  case shareCount > 1 && linkCount > 1:
    // many shares many links
    description = this.$gettext(
      "This %{ resourceType } is shared via %{ shareCount } invites and %{ linkCount } links"
    )
    break
}
```

kulmann: Still not happy with it. Pluralization works differently in different languages. Therefore we _must_  use `$ngettext`. But it's simply not possible to have multiple `n`s. A solution would be to have two independent sentences. One for links, one for invites. Doing proper plualization and variable injection in the full strings. And then just concatenating the fully translated independent sentences as one resulting string. I.e.

`$ngettext("This %{ resourceType } is shared via %{ linkCount } link", "This %{ resourceType } is shared via %{ linkCount } links", links.length)`.

Further reading: http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms

TL;DR: don't enforce pluralization from your own language. :trollface: Sorry for being a PITA with this.

pascalwengerter: Okay so I kinda reverted/refactored the changes before the `switch-case`-piece of art and rebased to master, let me know what you think about this one!

src/components/table/OcTableFiles.vue
```diff
@@ -388,6 +390,53 @@ export default {
...
+      const linkText =
+        linkCount > 0
+          ? this.$ngettext(
+              "This %{ resourceType } is shared via %{ linkCount } link",
+              "This %{ resourceType } is shared via %{ linkCount } links",
+              linkCount
+            )
+          : ""
+
+      const description = shareText + (shareCount > 0 && linkCount > 0 ? " " : "") + linkText
```
kulmann, line 418: Can be simplified to
```suggestion
      const description = [shareText, linkText].join(' ')
```

kulmann (approved):
Ok, I'll approve this now 😁 Suggested one tiny improvement. But LGTM 💯

src/components/table/OcTableFiles.vue
pascalwengerter: One linter and unit test failure later this should actually be 🆗 🤒  (we're rendering a space now even if there's only one of the two present, but that's fine I suppose)

kulmann: With my solution (the `join`) the space doesn't happen if only one element is present.

pascalwengerter: Tried it in the console (`test = ["one", ""].join(" ")` which returns `"one "`, also the snapshot test suggests it does so? Or am I missing sth?

kulmann: Ah. ups. Right 😅 You could add `trim()` to the result of `join(" ")`

PR #1251, opened 2021-04-29
Make ariaLabel in OcSpinner optional

pascalwengerter:


JammingBen (approved):
Haha nice, I did https://github.com/owncloud/owncloud-design-system/pull/1251, but this makes more sense as the spinner is inside an `aria-hidden` element 🤺

PR #1252, opened 2021-04-29
Make files table headings more descriptive

JammingBen:


JammingBen:
I'm not quite sure how to handle the requirement for the status column: `check if table column header makes sense for combined action+status`. @pascalwengerter As you participated in the workshop, any ideas?

JanAckermann (approved):

kulmann:
"Shared with" is/was already fine. Please keep in mind that voice over gives the header as context when reading a data cell. So "Shared with" as header and the "Albert Einstein, Marie Curie,
quick action link and 3 more" is already appropriate. No need to give more context of what kind of share receivers can be in the data cell.

kulmann:
> I'm not quite sure how to handle the requirement for the status column: `check if table column header makes sense for combined action+status`. @pascalwengerter As you participated in the workshop, any ideas?

The thing is that the share status column is two things at the same time: actions for accepting/declining shares and the current share state. The question is "what's more important, status or actions?" and give the name respectively. IMO it's not an option to split into two columns. 

There is a PR coming from Lisa from Cern which will split the table into three different tables. One for pending, one for accepted, one for declined. With that in mind I'd give the column a name that is about the action. But I can't think of a good one which is not too long.

JammingBen:
> There is a PR coming from Lisa from Cern which will split the table into three different tables. One for pending, one for accepted, one for declined. With that in mind I'd give the column a name that is about the action. But I can't think of a good one which is not too long.

Ahh okay. Hmmm. What's a bit confusing for me is that we have two action-columns then. One for the regular actions, one for accepting/declining shares.

I mean, `actions` seems to be a good heading overall. But with two action-columns, we might need to look for a proper wording on both columns! Otherwise I would wonder why the "accepting/declining shares"-action is not included in the "actions" column.

pascalwengerter:
Since the status/sharing state will be covered by getting sorted into a separate table (Elizaveta's PR) soon and we aren't under timepressure in terms of the a11y certification I'd vote in favor of keeping the "Actions" column name instead of changing it now and then changing it back soon(?)

kulmann (approved):

PR #1263, opened 2021-04-30
add oc-tooltip directive

fschade:
add own tooltip directive to replace uk-tooltip


pascalwengerter (approved):
LGTM 🔥

pascalwengerter (approved):
Codacy wants a minor code change but then we should be ready to merge!

PR #1264, opened 2021-04-30
Reduce resource component type to router-link

pascalwengerter:
Would like to discuss this with @kulmann on Monday before merging - I still see the (potential) need for rendering buttons as resource type, but right now it's an a11y offense. This PR is a quickfix, so I didn't cleanup the codebase, tests are still failing and perhaps there's a better solution, too


kulmann:
We also have `a` tags (e.g. opening a file in a new tab, or downloading a file). IMO not possible to reduce to `router-link` only.

pascalwengerter:
Closing this as outdated, we'll open another PR once it's clear how we can handle the topic in oCIS

PR #1268, opened 2021-05-03
Added l10n no options text, added translation for vue-select combobox aria-label

JanAckermann:
This PR addresses to issues:

1. Adding a default No options available slot, which appears if there are no items passed to the component or the while type in a search query without a result. This text will also be synced to transifex.

2. Unfortunately there is an aria label in the combo box with the text: "Search for option" which can't be replaced via the vue-select API, in order to sync the text to transifex we need to replace it. 
Screenshot from 2021-05-03 11-48-15

### DIFF
--- a/src/components/OcSelect.vue
+++ b/src/components/OcSelect.vue
@@ -3,7 +3,7 @@
     <template v-for="(index, name) in $scopedSlots" #[name]="data">
       <slot :name="name" v-bind="data"></slot>
     </template>
-    <template slot="no-options" v-translate>No options available.</template>
+    <div slot="no-options" v-translate>No options available.</div>
   </vue-select>
 </template>
 
@@ -159,9 +159,6 @@ We can then retrieve all the values that we want to display from the slots param
           <strong v-text="title" />
         </span>
         <span class="option" v-text="desc" />
-      </template>
-      <template #no-options>
-        Your search query haven't returned any results.
       </template>
     </oc-select>
     <p>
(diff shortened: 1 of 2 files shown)


src/components/OcSelect.vue
```diff
@@ -149,9 +160,6 @@ We can then retrieve all the values that we want to display from the slots param
         </span>
         <span class="option" v-text="desc" />
       </template>
-      <template #no-options>
```
LukasHirt, line 152: As it's still possible to include custom no options message via slot, keet this example pls.

src/components/OcSelect.vue
```diff
@@ -1,8 +1,9 @@
 <template>
-  <vue-select :value="model" class="oc-select" v-bind="$attrs" v-on="$listeners">
+  <vue-select ref="select" :value="model" class="oc-select" v-bind="$attrs" v-on="$listeners">
     <template v-for="(index, name) in $scopedSlots" #[name]="data">
       <slot :name="name" v-bind="data"></slot>
     </template>
+    <div slot="no-options" v-translate>No options available.</div>
```
LukasHirt, line 6: Make it possible to replace this via a slot

JanAckermann:
Created an issue(FR) on vue-select about localize aria-label on combo box in beautiful way 
https://github.com/sagalbot/vue-select/issues/1420

pascalwengerter (approved):
Great stuff 🙌🏽

PR #1269, opened 2021-05-03
replace uk-drop by using tippy

fschade:
this pr replaces uk-drop by using tippy (same as tooltip directive).

The oc-drop api was kept intact so no changes by the consuming app expect importing tippy as peer dependency should be required.

It also keeps care that a11y (accessibility) usage of the web-ui is fulfilled.

Because it's not a breaking change it can be part of the upcoming 6.4.0 release.

### DIFF
--- a/src/components/OcDrop.vue
+++ b/src/components/OcDrop.vue
@@ -106,6 +106,9 @@ export default {
       interactive: true,
       plugins: [hideOnEsc],
       theme: "none",
+      aria: {
+        content: "describedby",
+      },
       onShow(instance) {
         hideAll({ exclude: instance })
       },
--- a/src/components/OcDrop.vue
+++ b/src/components/OcDrop.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :id="dropId" class="oc-drop" :uk-drop="$_ocDrop_props" @click="$_ocDrop_close">
+  <div :id="dropId" ref="drop" class="oc-drop" @click="$_ocDrop_close">
     <div v-if="$slots.default" class="uk-card uk-card-default uk-card-small uk-card-body">
       <slot />
     </div>
@@ -7,7 +7,9 @@
   </div>
 </template>
 <script>
-import UiKit from "uikit"
+import tippy, { hideAll } from "tippy.js"
+
+import { destroy, hideOnEsc } from "../directives/OcTooltip"
 import uniqueId from "../utils/uniqueId"
 
 /**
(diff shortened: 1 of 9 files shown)


src/components/OcDrop.vue
```diff
@@ -26,29 +28,22 @@ export default {
...
      * The position of the drop.
      **/
     position: {
       type: String,
-      default: "bottom-left",
+      default: "bottom-start",
       validator: value => {
         return value.match(
-          /((top|bottom)-(left|center|right|justify)|(left|right)-(top|center|bottom))/
+          /((top|right|bottom|left|auto)|(top|right|bottom|left|auto)-(start|end))/
```
kulmann, line 46: Could you add a line to the docblock of this prop about the valid combinations? Like `Use top, right, bottom, left or auto for centered placement or add '-start'  or '-end'  as a suffix for altering the alignment`?
Also: is `auto-start` really a valid combination? 🤔

fschade: Yea it’s valid, but to be honest I haven’t tried the auto prop...

 // default
  placement: 'top',

  // full list:
  placement: 'top-start',
  placement: 'top-end',

  placement: 'right',
  placement: 'right-start',
  placement: 'right-end',

  placement: 'bottom',
  placement: 'bottom-start',
  placement: 'bottom-end',

  placement: 'left',
  placement: 'left-start',
  placement: 'left-end',

  // choose the side with most space
  placement: 'auto',
  placement: 'auto-start',
  placement: 'auto-end',

kulmann: Please don't add the full list. 😁 After all it's devdocs, so people reading it will be capable of understanding a hint about the combinations. Just the regex would be nice to avoid in the docs. ;-)

kulmann (commented):
What about focus management? At least here in the examples the `Filter` drop has checkboxes, so we'd probably need to give the focus to the drop as soon as it opens and move it back to the button when it closes. What do you think?

This was not the case before AFAIK, so I wouldn't block the PR with questions about focus management...

kulmann (approved):
👍

fschade:
> What about focus management? At least here in the examples the `Filter` drop has checkboxes, so we'd probably need to give the focus to the drop as soon as it opens and move it back to the button when it closes. What do you think?
> 
> This was not the case before AFAIK, so I wouldn't block the PR with questions about focus management...

missed your comment, sorry :( We keep care in the dedicated focus tasks.

PR #1272, opened 2021-05-04
Make tooltip aria-hidden and remove aria-expanded on calling element

pascalwengerter:


src/directives/OcTooltip.js
```diff
@@ -49,9 +61,11 @@ export default {
       ignoreAttributes: true,
       ...(Object.prototype.toString.call(value) === "[object Object]" && value),
       aria: {
-        content: "describedby",
+        content: null,
+        expanded: false,
+        hidden: true,
```
fschade, line 66: this can go because tippy doesn't support the option and we manage it our own by ariaHidden plugin

fschade (commented):
just one change, expect that 🚀

src/directives/OcTooltip.js
pascalwengerter: Thanks, should be resolved!

fschade (approved):
LGTM 🚀

PR #1282, opened 2021-05-05
Internalize vue-avatar to ODS, improve accessibility & support lazy img loading

pascalwengerter:


src/components/avatars/OcAvatar.vue
```diff
@@ -48,18 +64,126 @@ export default {
...
+     */
     width: {
       type: Number,
       required: false,
       default: 50,
     },
+    /**
+     * The initials get computed by the component itself
+     */
+    initials: {
```
kulmann, line 78: Why do we need the initials as a prop? From my understanding we can have the initials of the userName as a computed property. IMO the parser is not needed for us.

kulmann (changes requested):

src/components/avatars/OcAvatar.vue
pascalwengerter: Successfully re-re-rebased this and implemented the feedback, ready for re-review

src/components/OcImage.vue
```diff
@@ -36,6 +36,14 @@ export default {
       required: false,
       default: null,
     },
+    loadingType: {
```
kulmann, line 39: Loading type is not applied to the `img` tag in the template

src/components/avatars/OcAvatar.vue
```diff
@@ -1,31 +1,50 @@
...
+    class="vue-avatar--wrapper oc-avatar"
+    :style="[style]"
+    :width="width"
     :aria-label="accessibleLabel === '' ? null : accessibleLabel"
     :aria-hidden="accessibleLabel === '' ? 'true' : null"
     :focusable="accessibleLabel === '' ? 'false' : null"
     :role="accessibleLabel === '' ? null : 'img'"
-  />
+  >
+    <oc-img v-if="isImage" loading="lazy" class="avatarImg" :src="src" @error="onImgError" />
```
kulmann, line 11: prop name is not `loading`, but `loading-type`

src/components/avatars/OcAvatar.vue
```diff
@@ -1,31 +1,50 @@
 <template>
-  <avatar
-    class="oc-avatar"
-    :username="userName"
-    :size="width"
-    :src="src"
+  <span
+    class="vue-avatar--wrapper oc-avatar"
+    :style="[style]"
```
kulmann, line 4: Why are the styles wrapped into an array? 🤔

kulmann (changes requested):

src/components/avatars/OcAvatar.vue
pascalwengerter: Leftovers from the vue-avatar copy-pasta, thanks I'll get rid of it

src/components/OcImage.vue
pascalwengerter: Might've lost this one during the rebases, brining it back now 💪🏽

kulmann: ❤️

kulmann (approved):
Awesome! 🥇

PR #1291, opened 2021-05-10
Move tr height from prop to token

pascalwengerter:
First idea was to use existing padding/margin classes, however they don't properly work in table rows plus re-using the generic spacing classes to configure a more dense/lofty table row will yield unwanted results in the rest of the UI so I went for introducing a dedicated table row height CSS variable

### DIFF
--- a/src/components/table/OcTable.vue
+++ b/src/components/table/OcTable.vue
@@ -148,14 +148,6 @@ export default {
       default: null,
     },
     /**
-     * Height of the row in pixels
-     */
-    rowHeight: {
-      type: Number,
-      required: false,
-      default: 64,
-    },
-    /**
      * Top position of header used when the header is sticky in pixels
      */
     headerPosition: {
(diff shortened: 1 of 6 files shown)


src/components/table/OcTable.vue
```diff
@@ -39,7 +39,6 @@
     <tfoot v-if="$slots.footer" class="oc-table-footer">
       <tr>
         <td :colspan="footerColspan" class="oc-table-footer-cell">
-          <!-- @slot Footer of the table -->
```
kulmann, line 42: Why remove this line? It's used for the generated ODS docs. ;-)

pascalwengerter: Felt more like a stray line since it's the only comment in the component template (there's 3 slots inside), but happy to revert the deletion if wanted!

pascalwengerter: Reverted this line, let's merge this PR into master since the `6.0.2` release (branch) will be a dead end?

PR #1298, opened 2021-05-12
A11y improvements in avatar link & avatar count

### LINKED ISSUE #1287: a11y color contrast offense in OcAvatarLink
We'll need to update the way we handle background colors in the OcAvatarLink component since we need a color contrast ratio of >4.5 between background and font color

Remedy could be either using the OcAvatar component or porting the colors from there (which I made sure have enough contrast) to the OcAvatarLink

Screenshot_2021-05-07 Files shared via link - ownCloud

pascalwengerter:
Fixes #1287


fschade (approved):
LGTM 🚀

PR #1300, opened 2021-05-14
feat: rename sidebar to sidebar-nav, use fc and switch to slots instead of props

LukasHirt:
**Change: Rename OcSidebar to OcSidebarNav**

We've changed the name of the `OcSidebar` component to `OcSidebarNav` to better reflect that it shouldn't be used for anything else than the navigation.

**Change: Use slots in OcSidebarNav**

We've moved away from defining the nav items in the `OcSidebarNav` via props to including them though slots.
The component now contains 3 slots - `header`, `nav` and `footer`.

**Enhancement: Add inverse button variant**

We've added inverse button variant which should be used on a dark background.


src/components/OcSidebarNav.vue
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+</style>
+
+<docs>
+Sidebar navigation component containing branding
+
+## Examples
+```vue
+<oc-sidebar class="uk-height-large">
```
kulmann, line 79:
```suggestion
<oc-sidebar-nav class="uk-height-large">
```

src/components/OcSidebarNav.vue
```diff
@@ -0,0 +1,99 @@
...
+    <oc-list>
+      <oc-sidebar-nav-item icon="folder">Files</oc-sidebar-nav-item>
+      <oc-sidebar-nav-item :active="true" icon="shared-with-me">Shared with me</oc-sidebar-nav-item>
+      <oc-sidebar-nav-item icon="delete">Deleted files</oc-sidebar-nav-item>
+    </oc-list>
+  </template>
+  <template #footer>
+    <p class="oc-px-s">Made by ownClouders</p>
+  </template>
+</oc-sidebar>
```
kulmann, line 97:
```suggestion
</oc-sidebar-nav>
```

kulmann (changes requested):

LukasHirt:
@kulmann I applied your suggestions. Thanks for those catches!

kulmann (approved):
LGTM 👍

Issue #1289, opened 2021-05-07
Use OcSelect component for OcAutocomplete internally


kulmann:
We brought the `vue-select` component into our design system as `OcSelect` component for various reasons. Details can be found here: https://github.com/owncloud/owncloud-design-system/issues/762

Would be good to use the `OcSelect` now for the `OcAutocomplete` component internally as well.

Right now the `OcAutocomplete` component works with an input element for searching and a uikit `uk-drop` for the dropdown holding the items. The `OcSelect` component already brings along all of that. An example usage of the `OcSelect` component can be found here: https://github.com/owncloud/web/blob/8a2284fb7a0c8c3564874b72ec363d5c4773fe07/packages/web-app-files/src/components/Collaborators/CollaboratorsEditOptions.vue#L4

kulmann:
Positive side effect of this is, that we'd get rid of one more usage of UIkit.

pascalwengerter:
Closing this since we've deprecated the OcAutocomplete and will drop it in the future

PR #1301, opened 2021-05-17
WIP for #1289: Use OcSelect component for OcAutocomplete internally

paulneubauer:
Basic implementation has been done, still some points have to be added to complete issue 1289.

- Styling should be refactored because the dom hierarchy changed due using oc-select
- During delayed fetch loading spinner should be added again
- There are also changes made in the Oc-Select, so the last sample example implementation should be updated (using filter)
- It should be checked if aria is still implemented the right way (vue-select uses it as far as I know but still should be checked)

Thats all. Thanks for the challenge I really liked it :)


dschmidt:
The issue is mentioned in the title but not actually linked: https://github.com/owncloud/owncloud-design-system/issues/1289

kulmann:
Superseded by https://github.com/owncloud/owncloud-design-system/pull/1521

PR #1302, opened 2021-05-17
a11y color contrasts

fschade:
The contrasts of the color tokens are now a11y compliant. It was tested in latest oc-web with the Accessibility Insights for Web extension.

Additionally the contrast check rule in color tokens now supports to check multiple contrast rations with multiple backgrounds


pascalwengerter (changes requested):
Looks pretty neat, the code is quite complex though. Some notes:
- could you add the usage of this to the changelog item? E.g. "The color contrast checker now runs when generating tokens via the `yarn token` command and reports if any colors don't match the minimum color contrast"
- Coday requests some changes, could you tackle them or are they valid?
- Should we add some tests for some of the scripts at least?

fschade:
> Looks pretty neat, the code is quite complex though. Some notes:
> 
> could you add the usage of this to the changelog item? E.g. "The color contrast checker now runs when generating tokens via the `yarn token` command and reports if any colors don't match the minimum color contrast"

Good catch, done

> * Coday requests some changes, could you tackle them or are they valid?

the requested changes where wrong and we decided to disable it

> * Should we add some tests for some of the scripts at least?

as those scripts generating code only it's not required

kulmann (approved):
nice!

PR #1309, opened 2021-05-19
feat: add sort button to table header cells

LukasHirt:
I am still keeping the event handlers on the cell itself as well just to make it still clickable. This way we have solutions for both - click anywhere on the cell and button for keyboard nav.

image

Some additional changes:
- do not increase the size of table header via css prop same as normal rows
- align the button next to the text instead of end of the cell


pascalwengerter:
Checked this out and tried it locally, shouldn't we have the button span the whole `th` cell? Having to click on the (relatively small) icon feels cumbersome for me, but maybe there's other opinions/favors?

pascalwengerter:
Plus needs updated snapshot tests :)

LukasHirt:
> Having to click on the (relatively small) icon feels cumbersome for me

The TH element still has the click event so clicking on the span is not necessary. Or at least it shouldn't be. If it's not working for you then I'll try to debug it but might be hard as it worked for me :shrug:

pascalwengerter:
> > Having to click on the (relatively small) icon feels cumbersome for me
> 
> The TH element still has the click event so clicking on the span is not necessary. Or at least it shouldn't be. If it's not working for you then I'll try to debug it but might be hard as it worked for me 🤷

Well turns out you're right (ofc 😉) but I got confused since only the button got `cursor: pointer;` (the mouse "click" hand). Could you add it to the `<th>` hover?

changelog/unreleased/enhancement-use-btn-to-sort
```diff
@@ -0,0 +1,5 @@
+Enhancement: Use button to trigger table sort
+
+We've added a button to the table head cell which can trigger sort for its column so that it is keyboard accessible.
```
pascalwengerter, line 3: Please add the additional changes mentioned in the PR description here, too :)

pascalwengerter (changes requested):

changelog/unreleased/enhancement-use-btn-to-sort
pascalwengerter: Plus update snapshot tests to make CI happy, code-wise it looks good!

src/components/table/OcTable.spec.js
```diff
@@ -92,7 +91,6 @@ describe("OcTable", () => {
     })
 
     expect(wrapper.findAll("oc-thead-stub").length).toBe(0)
-    expect(wrapper).toMatchSnapshot()
```
LukasHirt, line 95: Screw 💩  code snapshots... the cons heavily outweigh the pros of using them... saying this as the guy who actually implemented the first one in this codebase 🙈

src/components/table/OcTable.vue
```diff
@@ -1,18 +1,35 @@
...
   <table v-bind="extractTableProps()">
     <oc-thead v-if="hasHeader">
-      <oc-tr>
+      <oc-tr class="oc-table-header-row">
         <oc-th
           v-for="(field, index) in fields"
           :key="`oc-thead-${field.name}`"
           v-bind="extractThProps(field, index)"
           @click.native="$emit(constants.EVENT_THEAD_CLICKED, field)"
           @keydown.enter.native="$emit(constants.EVENT_THEAD_CLICKED, field)"
```
kulmann, line 10: Could you remove the keydown on the `th`? Right now you can `TAB` through the table header row, focussing the `<th>` first and additionally the `<oc-button>` after, so you result in two focussable elements solving sorting, while you only need one (which should be the button for the keyboard, as it's the native element for that).

kulmann: Also, please remove the `props.tabindex = "0"` from the `sort.js` mixin so that the header cell is not focusable anymore. It's still correct that the aria-sort goes to the `th`.

kulmann (changes requested):
Awesome! Found one issue regarding keyboard navigation which needs easy-fixing-by-code-deletion. 😁

kulmann (approved):
😍

src/components/table/OcTable.spec.js
pascalwengerter: They have saved me more than once from committing crap so 🤷🏽 kudos to the original implementator!

pascalwengerter (approved):

PR #1310, opened 2021-05-19
Change OcSidebarNav active and hover colors

kulmann:
This PR switches background color for nav items in `hover` and `active` state within the `OcSidebarNav` component, now fulfilling required color contrasts for WCAG 2.1 AA.

Up for discussion if white is too heavy.

Screenshot 2021-05-19 at 21 59 20

Screenshot 2021-05-19 at 21 59 04


pascalwengerter (approved):
LGTM 🎯

PR #1313, opened 2021-05-20
apply color to every child element of oc-alert by default

fschade:
as title says, every child element of oc-alert now get's oc-color-text-inverse


LukasHirt:
Since you're touching the scss file already, could you pls also move all the styles directly to the SFC? Thanks!

fschade:
> Since you're touching the scss file already, could you pls also move all the styles directly to the SFC? Thanks!

done :)

kulmann (approved):
LGTM 🎨

PR #1315, opened 2021-05-21
Remove background color transition in sidebar nav

kulmann:
This PR removes the background color transition in the sidebar nav items, since it turned out to be too heavy now with high contrast colors.


pascalwengerter (approved):
LGTM 🎨

PR #1316, opened 2021-05-21
Fix hover svg fill in OcButton

kulmann:
This PR fixes that icons in OcButtons were not flipped to the inverse text color on hover. Again, the color transition was too harsh with high contrast, so I removed it.


pascalwengerter (approved):
LGTM 💟

PR #1320, opened 2021-05-24
feat: add pagination component

LukasHirt:


LukasHirt:
Not yet sure if it makes sense to handle splitting the resources in this component or already when passing down as a prop

kulmann:
> Not yet sure if it makes sense to handle splitting the resources in this component or already when passing down as a prop

IMO the pagintation component should only emit events. Splitting the resources according to the selected page (same for filtering and actually also sorting) needs to happen in web, so that we don't have to change anything in ODS when switching to server side pagination/filtering/sorting. If I remember correctly we sort in ODS atm. That should be shifted over to web at some point in the future.

LukasHirt:
> IMO the pagintation component should only emit events

That would be a bit complex since it should be links, wouldn't it? We'd need to prevent the default action when clicking on them and then emit those events... Of course, then take care of switching the routes via some methods in the Web or any extension instead of one central location.

But yeah, agreed on the splitting especially when thinking about the server limit

kulmann:
Mhh yes, you're right. Needs to be solved via routing because of a11y and because we already said that we want to have the pagination state in the route.

LukasHirt:
Closing in favour of placing the pagination directly in the Web

PR #1327, opened 2021-05-26
Aria-labelledby through h2 in modals

pascalwengerter:


src/components/OcModal.vue
```diff
@@ -11,7 +11,7 @@
       >
         <div class="oc-modal-title">
           <oc-icon v-if="icon" :name="icon" :variation="variation" />
-          <h2 v-text="title" />
+          <h2 id="oc-modal-label" v-text="title" />
```
kulmann, line 14: Why not `oc-modal-title` as id? 🤔

pascalwengerter: Also fine for me, saw the a11y example use `-label` and went with it..

kulmann: I'd prefer `oc-modal-title` 😁

pascalwengerter: `oc-modal-title` it is, then

kulmann (approved):
👍

PR #1329, opened 2021-05-27
Text input clear

kulmann:
Adding an optional clear button to the OcTextInput component. It has a default aria-label (`Clear input`) which can be overridden via a prop. Button can be reached via keyboard navigation and used via `ENTER` key or via click. Using the button clears the text input and emits an `input` event with an empty string.

Screenshot 2021-05-27 at 14 45 45
Screenshot 2021-05-27 at 14 45 56
Screenshot 2021-05-27 at 14 46 04


fschade (approved):
LGTM 🎸

PR #1334, opened 2021-05-28
Make active breadcrumb item focus-able

pascalwengerter:


src/components/OcBreadcrumb.vue
```diff
@@ -13,7 +13,7 @@
         >
           {{ item.text }}
         </oc-button>
-        <span v-else :aria-current="getAriaCurrent(index)" v-text="item.text" />
+        <span v-else :aria-current="getAriaCurrent(index)" tabindex="-1" v-text="item.text" />
```
LukasHirt, line 16: A negative value doesn't allow sequential keyboard navigation to the el. Is that what we want? If yes, I find the changelog a bit confusing as it says "make it focusable for keyboard users"

pascalwengerter: You're totally right it's a contradiction and the changelog item does not get the point across well. Fixing it, thanks for the sharp eye 🕵🏽

LukasHirt (approved):
LGTM :+1:

PR #1349, opened 2021-06-02
Unify syntax highlighting in docs examples

### LINKED ISSUE #734: Use JSX for examples
Currently, we are using the Vue SFC syntax in the examples. What this results in are annoying errors. What would you say about using JSX instead for the examples so we don't see the error? The syntax is not that much different.

SFC:
image

JSX:
image

What do you think @PVince81 @kulmann ?

pascalwengerter:
Closes #734

### DIFF
--- a/src/components/OCNotifications.vue
+++ b/src/components/OCNotifications.vue
@@ -42,7 +42,7 @@ export default {
 </script>
 
 <docs>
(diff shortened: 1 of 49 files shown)


kulmann (approved):
👍

PR #1352, opened 2021-06-03
Internalize some uikit classes to OC

pascalwengerter:
No changelog item since it (basically) only touches examples

### DIFF
--- a/src/components/OcSelect.vue
+++ b/src/components/OcSelect.vue
@@ -97,7 +97,7 @@ For detailed documentation (props, slots, events, etc.), please visit https://vu
 
 ```js
 <template>
-  <div class="uk-width-medium">
+  <div class="oc-docs-width-medium">
     <oc-select v-model="selected" :options="['Bannana', 'Orange', 'Pear']" />
   </div>
 </template>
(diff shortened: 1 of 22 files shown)


src/components/OcSelect.vue
```diff
@@ -97,7 +97,7 @@ For detailed documentation (props, slots, events, etc.), please visit https://vu
 
 ```js
 <template>
-  <div class="uk-width-medium">
+  <div style="width: 300px">
```
dschmidt, line 100: Not sure what your policies are but some people consider inline style to be the devil's work :trollface:

dschmidt: Is this a 1:1 replacement? or a change in behavior?

pascalwengerter: Only happens inside the `<docs>` examples but yeah I've moved it to a dedicated class within the docs' SCSS files now. It's a 1:1 replacement of setting a fixed width for the select box

dschmidt:
cool, lgtm - I'm just a bit reluctant to approve cause I'm not so familiar with this project/repo :)

pascalwengerter:
> cool, lgtm - I'm just a bit reluctant to approve cause I'm not so familiar with this project/repo :) 

No worries, thought it could be reasonable to slowly involve you some more since you're starting to use the components if I'm not mistaken 😉 thanks for taking the time!

kulmann (approved):
Cool 👍

PR #1364, opened 2021-06-10
[tests-only] Added unit tests for oc button

kiranparajuli589:
## Description
tests if
- button emits click event on click
- button is not clickable while disabled
- button can show provided slot

## Related Issue

- https://github.com/owncloud/web/issues/5217
- #1366 

## How Has This Been Tested?

- :robot:

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/OcButton.spec.js
+++ b/src/components/OcButton.spec.js
@@ -1,0 +1,193 @@
+import OcButton from "./OcButton.vue"
+import { createLocalVue, shallowMount } from "@vue/test-utils"
+const localVue = createLocalVue()
+
+/**
+ * There are some errors that should be caught but sneaked by vue.
+ * Vue Warn errors is one of them
+ *
+ * Global console contains these error from where we can assert if they are expected errors
+ */
+beforeEach(() => {
+  const { error } = global.console
+  global.console.error = (...args) => {
+    for (let i = 0; i < args.length; i += 1) {
+      const arg = args[i]
+      if (typeof arg === "string" && arg.includes("Vue warn")) {
+        // Vue warn contains messages with warning initiator element
+        // Only vue warn message is thrown
+        const errorSplit = arg.split("\n")
+        throw errorSplit[0]
+      }
+    }
+    error(...args)
+  }
+})
+
+describe("OcButton", () => {
+  const testSlots = { default: '<p class="text">Test button</p>' }
+  function getWrapperWithProps(props) {
+    return shallowMount(OcButton, { localVue: localVue, propsData: props })
+  }
+  function getWrapperWithTestSlot() {
+    return shallowMount(OcButton, { localVue: localVue, slots: testSlots })
+  }
+
+  it("should display slot html", () => {
+    const wrapper = getWrapperWithTestSlot()
+    const slot = wrapper.find("p")
+    expect(slot).toBeTruthy()
+    expect(slot.attributes("class")).toBe("text")
+    expect(slot.text()).toBe("Test button")
+  })
+
+  it("can stop class propagation", () => {
+    const wrapper = getWrapperWithProps({
+      stopClassPropagation: true,
+      size: "medium",
+    })
+    expect(wrapper.attributes("class")).toBe("")
+  })
+
+  describe("click event", () => {
+    it("should emit click event when click is triggered", async () => {
+      const wrapper = getWrapperWithProps({})
+      await wrapper.trigger("click")
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted("click")).toBeTruthy()
+    })
+    it.each`
+      type
+      ${"a"}
+      ${"router-link"}
+    `("should not emit click event when type is $type", async ({ type }) => {
+      const wrapper = getWrapperWithProps({ type: type })
+      await wrapper.trigger("click")
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted("click")).toBeFalsy()
+    })
+  })
+  describe("when oc button is disabled", () => {
+    let wrapper
+    beforeEach(() => {
+      wrapper = getWrapperWithProps({ disabled: true })
+    })
+    it("should have disabled attribute set to disabled", () => {
+      expect(wrapper.attributes("disabled")).toBe("disabled")
+    })
+    it("should not emit click event", async () => {
+      await wrapper.trigger("click")
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted("click")).toBeFalsy()
+    })
+  })
+  describe("different types of button", () => {
+    it.each`
+      type             | expectLink | expectButton | expectRouterLink
+      ${"a"}           | ${true}    | ${false}     | ${false}
+      ${"button"}      | ${false}   | ${true}      | ${false}
+      ${"router-link"} | ${false}   | ${false}     | ${true}
+    `("can behave as a $type", ({ type, expectLink, expectButton, expectRouterLink }) => {
+      const wrapper = getWrapperWithProps({ type: type })
+      expect(wrapper.find("a").exists()).toBe(expectLink)
+      expect(wrapper.find("button").exists()).toBe(expectButton)
+      expect(wrapper.find("router-link-stub").exists()).toBe(expectRouterLink)
+    })
+  })
+  describe("different sizes of button", () => {
+    it.each`
+      size         | expectedClass
+      ${"small"}   | ${"oc-button-s"}
+      ${"medium"}  | ${"oc-button-m"}
+      ${"large"}   | ${"oc-button-l"}
+      ${"x-small"} | ${"oc-button-undefined"}
+    `(
+      "when size prop is set as $size class $expectedClass should be assigned",
+      ({ size, expectedClass }) => {
+        const wrapper = getWrapperWithProps({
+          size: size,
+        })
+        expect(wrapper.attributes("class")).toContain(expectedClass)
+      }
+    )
+  })
+  describe("default prop values", () => {
+    /* eslint-disable no-unused-vars */
+    it.each`
+      name                 | expected
+      ${"size"}            | ${"oc-button-m"}
+      ${"variation"}       | ${"oc-button-passive"}
+      ${"justify content"} | ${"oc-button-justify-content-center"}
+      ${"gap size"}        | ${"oc-button-gap-m"}
+      ${"appearance"}      | ${"oc-button-passive-outline"}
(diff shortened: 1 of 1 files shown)


individual-it (approved):

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,53 @@
+import OcButton from "./OcButton.vue";
+import {createLocalVue, shallowMount} from "@vue/test-utils";
+const localVue = createLocalVue()
+
+
+describe("oc button", () => {
```
kulmann, line 6: Please use the actual component name here:
```suggestion
describe("OcButton", () => {
```

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  const testSlots = { default: '<p class="text">Test button</p>' }
+  function getWrapperWithProps(props) {
+    return shallowMount(OcButton, { localVue: localVue, propsData: props })
+  }
+  function getWrapperWithTestSlot() {
+    return shallowMount(OcButton, { localVue: localVue, slots: testSlots })
+  }
+
+  describe("emit click event", () => {
+    it("emits click event when click is triggered", () => {
```
kulmann, line 16: Please use `async` and `await` as much as possible. This means:
- `async () => {` instead of `() => {`
- `await wrapper.vm.$nextTick()`
- the `expect` statement belongs on the top level of the function then

kulmann: Same for the other tests

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,53 @@
...
+      const wrapper = getWrapperWithProps({})
+      wrapper.trigger("click")
+      wrapper.vm.$nextTick(() => {
+        expect(wrapper.emitted('click')).toBeTruthy()
+      })
+    })
+    it("does not emits click event when disabled", () => {
```
kulmann, line 23:
```suggestion
    it("does not emit click event when disabled", () => {
```
(typo)

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,53 @@
...
+      const wrapper = getWrapperWithProps({
+        disabled: true
+      })
+      wrapper.trigger("click")
+      wrapper.vm.$nextTick(() => {
+        expect(wrapper.emitted('click')).toBeFalsy()
+      })
+    })
+    it("does not emits click event when input is triggered", () => {
```
kulmann, line 32: This test doesn't make sense to me. Can you elaborate why this would be a good idea?

kulmann (changes requested):

kulmann:
Feel free to group tests that belong together in a deeper nesting. E.g. you could add another `describe` level, like this:
```js
describe("OcButton", () => {
  describe("disabled", () => {
    ... all the scenarios that  are about disabled buttons
```

src/components/OcButton.spec.js
kiranparajuli589: if a slot has an input, will the wrapper be confused? I added this just to verify the emit for invalid events. I can remove it if this makes no sense

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,146 @@
...
+  describe("default prop values", () => {
+    it.each`
+      name                 | expected
+      ${'size'}            | ${'oc-button-m'}
+      ${'variation'}       | ${'oc-button-passive'}
+      ${'justify content'} | ${'oc-button-justify-content-center'}
+      ${'gap size'}        | ${'oc-button-gap-m'}
+      ${'appearance'}      | ${'oc-button-passive-outline'}
+    `
+    ('should have $name as $expected', ({name, expected}) => {
```
individual-it, line 141: name is not used :thinking:

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,146 @@
...
+        type: "router-link",
+        to: {
+          name: 'some-route'
+        },
+        disabled: true
+      })
+      await wrapper.trigger("click")
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted('click')).toBeFalsy()
+      console.log(wrapper.vm.$route.path)
```
individual-it, line 86: left over log output?

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,146 @@
...
+    it("does not emit click event", async () => {
+      await wrapper.trigger("click")
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted('click')).toBeFalsy()
+    })
+    it("should not route to provided router link", async () => {
+      wrapper = getWrapperWithPropsAndMockRouter({
+        type: "router-link",
+        to: {
+          name: 'some-route'
```
individual-it, line 79: if you use `route.path` in other places you could also use `route.name` here
or you hard-code it everywhere. As this is a property that is valid only for this single test a hard-coded value might make it more readable. And changing all routes in the whole test-file should not be ever needed

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,146 @@
...
+        },
+        disabled: true
+      })
+      await wrapper.trigger("click")
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted('click')).toBeFalsy()
+      console.log(wrapper.vm.$route.path)
+      expect(wrapper.vm.$route.path).toBe(route.path)
```
individual-it, line 87: see comment about route.path and route.name

src/components/OcButton.spec.js
kiranparajuli589: name is used in the description

individual-it (changes requested):
please use `' ` instead of `"`

src/components/OcButton.spec.js
kiranparajuli589: it is used in two places, so I've used `route.path` to avoid duplications

kiranparajuli589: i've removed the tests to check for routes because they're not working properly.  for this scenario, while prop `disable: true`, the route should not be updated but the current configuration is not catching that. I'll add route tests separately after some research

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    let wrapper
+    beforeEach(() => {
+      wrapper = getWrapperWithProps({})
+    })
+    it('emits click event when click is triggered', async () => {
+      await wrapper.trigger('click')
+      await wrapper.vm.$nextTick()
+      expect(wrapper.emitted('click')).toBeTruthy()
+    })
+    it('does not emit click event when input is triggered', async () => {
```
kulmann, line 44: How would you ever trigger an `input` on a button? 🤔

src/components/OcButton.spec.js
kulmann: 👍

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      ${'a'}           | ${true}    | ${false}     | ${false}
+      ${'button'}      | ${false}   | ${true}      | ${false}
+      ${'router-link'} | ${false}   | ${false}     | ${true}
+    `
+    ('can behave as a $type', (
+      {type, expectLink, expectButton, expectRouterLink}) => {
+        const wrapper = getWrapperWithProps({ type: type })
+        expect(wrapper.find('a').exists()).toBe(expectLink)
+        expect(wrapper.find('button').exists()).toBe(expectButton)
+        expect(wrapper.find('router-link-stub').exists()).toBe(expectRouterLink)
```
individual-it, line 76: click events should only be emitted for buttons https://github.com/owncloud/owncloud-design-system/blob/master/src/components/OcButton.vue#L180-L182
you could add a test for that here

individual-it (changes requested):
there is logic to set the appearance https://github.com/owncloud/owncloud-design-system/blob/master/src/components/OcButton.vue#L151-L158
would be good to test that too

I think the other props are also worth to test, including the validators, those are the public APIs of the button, so should be tested

src/components/OcButton.spec.js
individual-it: I would have put it all in a normal `it`
with multiple `expect` in it, but OK - if you like that style its good ;-)

src/components/OcButton.spec.js
kiranparajuli589: i've removed, this is making confusions

src/components/OcButton.spec.js
kiranparajuli589: tests added

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,181 @@
...
+    for (let i = 0; i < args.length; i += 1) {
+      const arg = args[i];
+      if (typeof arg === 'string'
+        && arg.includes('Vue warn'))
+      {
+        const errorSplit = arg.split("\n")
+        throw errorSplit[0]
+      } else if (typeof arg === 'string'
+          && arg.includes('unhandledRejection')) {
+        // throw new Error(format(...args));
```
individual-it, line 18: left overs?

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,181 @@
+import OcButton from './OcButton.vue';
+import {createLocalVue, shallowMount} from '@vue/test-utils';
+const localVue = createLocalVue()
+
+// this call will be set before every test
```
individual-it, line 5: I think it's clear what `beforeEach` does, a better comment would be to explain what you are trying to achieve here

src/components/OcButton.spec.js
```diff
@@ -0,0 +1,181 @@
...
+      ${'appearance'}      | ${'oc-button-passive-outline'}
+    `
+    ('should have attribute "$name" as "$expected"', ({name, expected}) => {
+      const wrapper = getWrapperWithProps({})
+      expect(wrapper.attributes('class')).toContain(expected)
+    })
+  })
+  describe("invalid prop value", () => {
+    it.each`
+      prop                | value
```
individual-it, line 130: why the second column of the table if all values are the same?

individual-it (changes requested):

individual-it (approved):

individual-it (approved):

individual-it (approved):

PR #1377, opened 2021-06-11
Fix/improve table sorting

### LINKED ISSUE #1370: OcTableFiles: Sorting is no working by clicking on "Share owner" column header
### Steps to reproduce
1. Click on the header "Share owner"

### Expected behaviour
The table is sorted in alphabetical ascending or (by the second click) descending order by share owner's surname as the the default user icons also show the first letter of the surname

### Actual behaviour
Clicking has no effect

pascalwengerter:
## Description
Currently, same sized arrays containing objects don't get properly sorted. We've added the capability to sort them based on their keys by passing a string or function

## Related Issue
- Fixes #1370

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [X] Code changes

## Open tasks:
- [X] Unit tests added

### DIFF
--- a/src/components/table/mixins/sort.js
+++ b/src/components/table/mixins/sort.js
@@ -28,11 +28,7 @@ export default {
         const { sortable } = this.fields.find(f => f.name === this.sortBy)
 
         if (sortable) {
-          if (
-            typeof sortable === "string" &&
-            Array.isArray.call(aValue) &&
-            Array.isArray.call(bValue)
-          ) {
+          if (typeof sortable === "string") {
             const genArrComp = vals => {
               return vals.map(val => val[sortable]).join("")
             }
--- a/src/components/table/mixins/sort.js
+++ b/src/components/table/mixins/sort.js
@@ -21,9 +21,29 @@ export default {
       }
 
       return [...this.data].sort((a, b) => {
-        const aValue = a[this.sortBy]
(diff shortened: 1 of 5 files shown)


src/components/table/mixins/sort.js
```diff
@@ -21,10 +21,26 @@ export default {
...
+
+        if (sortable) {
+          if (typeof sortable === "string") {
+            const genArrComp = vals => {
+              return vals.map(val => val[sortable]).join("")
+            }
+
+            aValue = genArrComp(aValue)
+            bValue = genArrComp(bValue)
+          } else if (typeof sortable === "function") {
```
fschade, line 38: can we add one test for function sortable?

pascalwengerter: Agreed that we should, do you have any idea of a use case?

kulmann (approved):
👍

src/components/table/mixins/sort.js
fschade: Maybe a star ⭐️ rating ?

pascalwengerter: The branch is merged but I raised an issue [here](https://github.com/owncloud/owncloud-design-system/issues/1437) if you want to get funky :)

PR #1384, opened 2021-06-15
Make OcAppSideBar header sticky & cleanup component

pascalwengerter:
## Description
Make OcAppSideBar header sticky

## Related Issue
- Fixes a ticket in `web` (and will need test selector updates there 😈  )

## Motivation and Context
See changelog item

### DIFF
--- a/src/components/OcAppSideBar.vue
+++ b/src/components/OcAppSideBar.vue
@@ -1,30 +1,24 @@
 <template>
   <div class="oc-app-side-bar">
-    <div class="sidebar-container">
-      <div class="header primary">
-        <div v-if="!disableAction" class="action">
-          <slot name="action">
-            <oc-button
-              class="uk-float-right"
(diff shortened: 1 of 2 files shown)


src/components/OcAppSideBar.vue
```diff
@@ -64,6 +58,13 @@ export default {
 .oc-app-side-bar {
   height: 100vh;
 }
+
+.oc-app-side-bar-header {
+  background-color: var(--oc-color-background-default);
+  position: sticky;
+  top: 0px;
```
LukasHirt, line 65: Sticky position needs the actual top position i.e. in web under the top bar...

LukasHirt: With the current state this can be ignored as we do not do "real" scrolling of the window but scroll containers instead

LukasHirt (approved):

PR #1388, opened 2021-06-16
Add breakpoints to theming initialization

pascalwengerter:
## Description
Technically a bugfix but since we haven't communicated the theme-ability of breakpoints it can count as a feature 👹 

## Motivation and Context
We have breakpoint CSS props and the ability to set custom CSS props via theming but were missing out on initializing (possibly present) breakpoint variables from a theming config passed by a ODS user


LukasHirt:
Something to keep in mind - css custom props cannot be used in media queries as those are not part of the `:root`

LukasHirt (approved):

PR #1393, opened 2021-06-17
show '--' for negative file size

### LINKED ISSUE #1402: [QA] show 'pending' when the folder size is not yet computed
New folders do not show a size.

Expected behaviour:
Gui elements should not go missing unexpectedly.
New folders should show an indicator when the computation is pending. 

Suggestion: Show  a '?' where the file size should be. Give explanation about pending, file scan, etc. as a mouse over tooltip.

individual-it:
## Description
1. use toEqual to assert content, because `toMatch` is expecting a regex, to `toMatch("")` would always be true, I noticed it after I changed the code and the test still passed
2. as discussed in #1402 show negative number as `--` and not empty string

## How Has This Been Tested?
unit tests

## Types of changes

- [x] Bug fix, fixes #1402
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-show-dashes-when-negative-filesize-given
+++ b/changelog/unreleased/bugfix-show-dashes-when-negative-filesize-given
@@ -1,0 +1,6 @@
+Bugfix: show `--` as file-size when the folder size is not yet computed
+
+When the folder size is unknown (not yet calculated or cannot be calculated) the server returns `-1` as the size. In this case we now show `--` and not only an empty string
+
+https://github.com/owncloud/owncloud-design-system/issues/1402
+https://github.com/owncloud/owncloud-design-system/pull/1393
--- a/src/components/resource/OcResourceSize.spec.js
+++ b/src/components/resource/OcResourceSize.spec.js
@@ -10,7 +10,7 @@ describe("OcResourceSize", () => {
       },
     })
 
-    expect(wrapper.find(".oc-resource-size").text()).toMatch("?")
+    expect(wrapper.find(".oc-resource-size").text()).toEqual("?")
   })
 
   it("shows an empty string for values smaller than 0", () => {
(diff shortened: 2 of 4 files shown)


LukasHirt:
Isn't a negative value displayed if the size is being still calculated?

individual-it:
How? 'size' should never be negative now

pascalwengerter:
> How? 'size' should never be negative now

I think he (and me as well) were wondering how a server can even return a negative value in the first place ;D

LukasHirt:
https://central.owncloud.org/t/understanding-pending-state/6884/3

> A folder appears pending if their size is "-1" in oc_filecache. It usually means that its content hasn't been scanned so the folder size is unknown since the contents size itself is also unknown.

individual-it:
@LukasHirt thank you for the explanation, I actually did not knew that.

individual-it:
should we show an `?` if the filesize is `-1`?

LukasHirt:
> should we show an ? if the filesize is -1?

Good question... I think this is how oC classic is handling this case so at least for now I'd say yes, let's do so

pascalwengerter:
FYI found this in web: https://github.com/owncloud/web/issues/2131

Not sure if `pending` or `?` are better @kulmann

pascalwengerter:
> FYI found this in web: [owncloud/web#2131](https://github.com/owncloud/web/issues/2131)
> 
> Not sure if `pending` or `?` are better, we may want to add context for a11y reasons? @kulmann

Actually transferred this issue to the ODS so this PR will close it 💪🏽

individual-it:
I've changed it to be `--` as proposed in #1402

pascalwengerter (approved):
LGTM 💯 love it

kulmann (approved):
🚀

Issue #1366, opened 2021-06-10
Add unit tests to first batch of components


kiranparajuli589:
Add unit tests for the following components:
- [x] OcAvatar.vue (#1424)
- [x] OcAvatarLink.vue (#1430)
- [x] OcAccordionItem.vue (#1381)
- [x] OcAccordion.vue (#1389)
- [x] OcActionDrop.vue (#1394)
- [x] OcAlert.vue (#709)
- ~OcAppBar.vue~ (currently unused and out of scope for this ticket)
- ~OcAppLayout.vue~ (currently unused and out of scope for this ticket)
- [x] OcAppSideBar.vue (#1410)
- [x] OcAutocomplete.vue (#1440)
- [x] OcBreadcrumb.vue (#806)
- [x] OcButton.vue (#1364)
- [x] OcCheckbox.vue (#1408)
- [X] OcDatepicker.vue~ (#1661)
- ~OcDisclosureDrop.vue~ (currently unused and out of scope for this ticket)
- [x] OcDrop.vue (#1269)
- [x] OcDropzone.vue (#1439)
- [x] OcGrid.vue (#1441)
- [x] OcHiddenAnnouncer.vue (#1445)
- [x] OcIcon.vue (#1407)
- [x] OcImage (#1446)
- ~OcList.vue~ (currently unused and out of scope for this ticket)
- [x] OcLoader.vue (#1449)
- [x] OcLogo.vue (#1450)
- [x] OcModal.vue (#723)
- [x] _OcNavbarItem.vue (#1451)
- [x] OcNavbar.vue (#1452)
- [x] OcNavItem.vue (#1475)
- [x] OcNav.vue (#1474)
- [x] OcNotificationMessage.vue (#1476)
- [x] OcNotifications.vue (#1489)
- ~OcProgressPie.vue~ (currently unused and out of scope for this ticket)
- [x] OcProgress.vue (#819)
- [x] OcRadio.vue (#1490)
- [x] OcSearchBar.vue (#1495)
- [ ] OcSelect.vue
- ~_OcSidebarNavDivider.vue~ (deprecated component)
- ~_OcSidebarNavHeader.vue~ (deprecated component)
- [ ] OcSidebarNavItem.vue
- [x] OcSidebarNav.vue (#1300)
- [ ] OcSpinner.vue
- [ ] OcStatusIndicators.vue
- ~OcSwitch.vue~ (currently unused and out of scope for this ticket)
[...]

PR #1394, opened 2021-06-17
Added unit tests for OcActionDrop

kiranparajuli589:
## Description
Added following uni tests:
- toggle dropdown
    - when toggle button is toggled
    - when dropdown is open and  menu is clicked
- open menu and focus

## Related Issue

- Part of #1366 

## How Has This Been Tested?

- 🤖 

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/OcActionDrop.spec.js
+++ b/src/components/OcActionDrop.spec.js
@@ -1,0 +1,68 @@
+import OcActionDrop from "./OcActionDrop.vue"
+import { mount } from "@vue/test-utils"
+
+describe("OcActionDrop", () => {
+  describe("toggle dropdown", () => {
+    let wrapper
+    beforeEach(async () => {
+      wrapper = await mount(OcActionDrop, {
+        slots: {
+          actions:
+            "<p class='slot-action'>First action</p>" + "<p class='slot-action'>Second action</p>",
+        },
+      })
+    })
+    it("should set dropdown menu as hidden at first", () => {
+      expect(wrapper.find(".oc-dropdown-menu").exists()).toBeTruthy()
+      expect(wrapper.find(".oc-dropdown-wrapper button").attributes("aria-expanded")).toBe("false")
+      expect(wrapper.find(".oc-dropdown-menu").attributes("hidden")).toBe("hidden")
+    })
+    describe("when toggle button is toggled", () => {
+      it("should toggle visibility of dropdown menu", async () => {
+        await wrapper.find(".oc-dropdown-wrapper button").trigger("click")
+        expect(wrapper.find(".oc-dropdown-menu").attributes("hidden")).toBe(undefined)
+        expect(wrapper.find(".oc-dropdown-wrapper button").attributes("aria-expanded")).toBe("true")
+        await wrapper.find(".oc-dropdown-wrapper button").trigger("click")
+        expect(wrapper.find(".oc-dropdown-menu").attributes("hidden")).toBe("hidden")
+        expect(wrapper.find(".oc-dropdown-wrapper button").attributes("aria-expanded")).toBe(
+          "false"
+        )
+      })
+    })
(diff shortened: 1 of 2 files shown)


src/components/OcActionDrop.vue
```diff
@@ -60,20 +60,25 @@
...
+  },
   props: {
     /**
      * Supplies button variation to give additional meaning.
      * `primary, danger`
      */
     buttonVariation: {
       type: String,
-      default: "default",
+      default: "passive",
```
kiranparajuli589, line 79: default is not a valid prop-value for oc-button variation prop

individual-it: how can we add tests, so that it does not happen again?

kiranparajuli589: validator used here and in the `OcButton` component will throw `vue [warn]` if the invalid prop is used.
we've to catch the console warning and test if waring for invalid props are not present, like in https://github.com/owncloud/owncloud-design-system/pull/1364/files where invalid prop vue warns are thrown as real error -> after merge we could use that before every test cases

individual-it (commented):
LGTM, but would like to have a review from @kulmann @LukasHirt or @pascalwengerter because of the changes in the code

src/components/OcActionDrop.vue
kulmann: Could you also add all other valid button variations? can be found here https://github.com/owncloud/owncloud-design-system/blob/64d5456127a30a6db5744cae9d0718fb1875476d/src/components/OcButton.vue#L93

src/components/OcActionDrop.spec.js
```diff
@@ -0,0 +1,68 @@
...
+        expect(wrapper.find(".oc-dropdown-menu").attributes("hidden")).toBe(undefined)
+        expect(wrapper.find(".oc-dropdown-wrapper button").attributes("aria-expanded")).toBe("true")
+        await wrapper.find(".oc-dropdown-wrapper button").trigger("click")
+        expect(wrapper.find(".oc-dropdown-menu").attributes("hidden")).toBe("hidden")
+        expect(wrapper.find(".oc-dropdown-wrapper button").attributes("aria-expanded")).toBe(
+          "false"
+        )
+      })
+    })
+    describe("when dropdown is open and  menu is clicked", () => {
```
kulmann, line 32:
```suggestion
    describe("when dropdown is open and menu is clicked", () => {
```

kulmann (changes requested):

src/components/OcActionDrop.vue
kiranparajuli589: fixed

saw-jan (commented):
LGTM

src/components/OcActionDrop.spec.js
```diff
@@ -0,0 +1,68 @@
...
+          data() {
+            return {
+              isOpen: true,
+            }
+          },
+        })
+        wrapper.vm.$refs.button.focus = jest.fn()
+      })
+      it("should close the dropdown menu", async () => {
+        await wrapper.find(".oc-dropdown-menu").trigger("click")
```
individual-it, line 45: maybe check first if the dropdown is open, to make sure we don't have a snake oil test here

individual-it (changes requested):
should we put `.oc-dropdown-menu` `.oc-dropdown-wrapper button` into a const?

src/components/OcActionDrop.spec.js
kiranparajuli589: fixed

individual-it (approved):

Issue #1357, opened 2021-06-07
Discussion: Rearrange CI steps


pascalwengerter:
### Expected behaviour
CI should fail as early as possible

### Actual behaviour
Unit tests are run as the last step, after building the docs & package. From my understanding that order isn't necessary, but maybe I'm mistaken?

### Screenshots

<img width="142" alt="Screenshot 2021-06-07 at 16 54 00" src="https://user-images.githubusercontent.com/16822008/121051089-11ff1280-c7b1-11eb-8479-e582185e5a28.png">

kulmann:
Additionally we now have the `Sonarcloud` pipeline. IMO some pipelines can run in parallel, to speed things up.

phil-davis:
ToDo:
- see which steps are really dependent on each other
- split into a couple of pipelines, as long as that reduces overall run-time
- do "dumb-easy" checks as early as possible (e.g. lint) so that a pipeline will fail early if those sort of checks are failing

pascalwengerter:
Hey @dpakach, thanks for taking care of this! From what I see in the last CI run this doesn't implement the idea I had in mind when writing the ticket though 😉 

What I originally wanted was to re-arrange steps so the "costly" build steps happen _after_ linting and tests have passed successfully. 

A rough idea was to have the following order:

- Clone
- Dependencies
- eslint
- stylelint
- unit tests
- sonarcloud
- build docs
- build system
- changelog/publish

How many steps this is separated into is kinda up to you

kulmann:
There are some things that can be parallelized, like e.g. `build docs` and `build system` can run in parallel. But those should only start if all the cheaper/easier steps were successful, like @pascalwengerter already explained.

dpakach:
@pascalwengerter  that sounds good too, I just thought that if we separate into different pipeline, we can just run the testing+linting and build pipelines in parallel which will also save a minute or two.
I dont think we can run the drone steps in parallel. There is a option for `detach` which can be set so that the steps behave like services. But the detached steps dont fail the build.

https://docs.drone.io/pipeline/exec/syntax/steps/#detach

kulmann:
Sorry, what I meant was having `build docs` and `build system` in their own pipelines respectively, and then let them depend on whatever makes sense to run before (the `cheap` pipelines)

pascalwengerter:
[...]

PR #1415, opened 2021-06-23
[ci] rearrange ci steps to optimize runtime

dpakach:
## Description

Break the testing pipeline step into lint, build and testing pipelines so that the CI can fail faster.

## Related Issue

- Fixes https://github.com/owncloud/owncloud-design-system/issues/1357

## How Has This Been Tested?

:robot:

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,15 +1,62 @@
 def main(ctx):
     before = [
+        build(ctx),
         testing(ctx),
     ]
 
     stages = [
         changelog(ctx),
-        build(ctx),
+        buildAndPublish(ctx),
     ]
 
     return before + stages
 
+
+def build(ctx):
+    return {
+        'kind': 'pipeline',
+        'type': 'docker',
+        'name': 'build',
+        'platform': {
+            'os': 'linux',
+            'arch': 'amd64',
+        },
+        'steps': [
+            {
+                'name': 'dependencies',
+                'image': 'owncloudci/nodejs:14',
+                'pull': 'always',
+                'commands': [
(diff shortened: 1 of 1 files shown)


kiranparajuli589:
IMO `lint` steps can also be separated from the test pipe to separate one 🤔

dpakach:
> IMO `lint` steps can also be separated from the test pipe to separate one  if fails on lint, do not run tests

the problem is that we need to install dependencies for all steps, and since the lint pipeline doesn't take much time (20 - 30 seconds), it is not required IMO

phil-davis (approved):
LGTM - this leaves the lint and unit tests in a separate pipeline that can run and fail early if there is a lint or unit test failure. The "build docs" and "build system" steps run in a separate pipeline that does not delay the unit test result....

It looks like it saves a couple of minutes elapsed time for a successful result.

pascalwengerter (changes requested):
Wait a sec

phil-davis:
Sorry - I  pressed merge!

LukasHirt:
@dpakach Why getting rid of `depends_on` in build steps? It actually makes those steps run in parallel so saving ~2mins per run.

pascalwengerter:
> Sorry - I pressed merge!

All good, opened a PR to revert this commit and commented in the (re-opened) ticket what I originally had in mind

PR #1420, opened 2021-06-24
Add three-dot/contextmenu slot to OcTableFiles

pascalwengerter:
## Description
See changelog item

## Related Issue
- Works towards https://github.com/owncloud/web/issues/5102
- Works towards https://github.com/owncloud/web/issues/5103

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] Code changes
- [x] Unit tests added

Spotted an opportunity of removing some `await nextTick` stuff from other tests, test suite seems happy about it and no user facing change hence no changelog item


.drone.star
```diff
@@ -51,35 +51,31 @@ def testing(ctx):
             },
             {
                 'name': 'dependencies',
-                'image': 'owncloudci/nodejs:14',
-                'pull': 'always',
+                'image': 'owncloudci/nodejs:16',
```
LukasHirt, line 54: v16 won't be LTS for next 3 months... I'd suggest to stay with v14 till then

pascalwengerter: Yeah I remember we had this discussion before 🥴 but since I want to use `.replaceAll()` which isn't supported in v14, but was supported in the now unsupported v15 and is supported in v16...idk, man

LukasHirt: You can use `.replace(regexp, ...)` to replace all.

src/components/table/helpers/constants.js
```diff
@@ -1,3 +1,4 @@
 export const EVENT_THEAD_CLICKED = "thead-clicked"
 export const EVENT_TROW_CLICKED = "highlight"
+export const EVENT_TROW_CONTEXTMENU = "contextmenuClicked"
```
fschade, line 3: EVENT_THEAD_CLICKED is kebab case, would it make sense to unify the style?

src/components/table/OcTableFiles.vue
```diff
@@ -337,6 +368,33 @@ export default {
...
   methods: {
+    resetDropPosition(id, event) {
+      const instance = this.$refs[id].tippy
+      this.displayPositionedDropdown(instance, event)
+    },
+
+    showContextMenu(rows, event) {
+      event.preventDefault()
+
+      const instance = rows.$el.getElementsByClassName("oc-table-files-btn-action-dropdown")[0]
```
fschade, line 379: is the element always available or shouldn't we take care with an early return?

src/components/table/OcTableFiles.vue
```diff
@@ -337,6 +368,33 @@ export default {
     },
   },
   methods: {
+    resetDropPosition(id, event) {
+      const instance = this.$refs[id].tippy
```
fschade, line 372: same as line 379

src/components/table/OcTableFiles.spec.js
```diff
@@ -168,4 +171,19 @@ describe("OcTableFiles", () => {
 
     expect(wrapper.emitted().fileClick[0][0].name).toMatch("forest.jpg")
   })
+
+  it("emitting contextmenuClicked event on right click on table row", async () => {
+    const tableRow = await wrapper.find(".oc-tbody-tr-forest")
```
kulmann, line 176: This test is not triggering a right click, so it doesn't match the description. I'd say either remove it (less coverage) or really trigger a right click.

fschade (approved):
some small comments, expect that LGTM 🎀

src/components/table/OcTableFiles.vue
kulmann: 👍 for early return. just to be sure.

.drone.star
pascalwengerter: Resolving this, thanks for the insights 🙌🏽

src/components/table/OcTableFiles.vue
kulmann: Hmmm... thinking about this now, wouldn't it make more sense to do something like this?
```js
const instances = rows.$el.getElementsByClassName("oc-table-files-btn-action-dropdown")
if (instances.length === 0) {
  return
}
```
?

kulmann: And then use `instances[0]` in line 387.

pascalwengerter: Very true yeah, fixing it

kulmann: Screenshot 2021-07-22 at 12 35 36

I should've read the spec, sorry for the noise. Your previous solution was perfectly fine, and shorter. 🙄

kulmann (approved):
LGTM 💪 🐭

PR #1421, opened 2021-06-24
feat: improve switch UI and a11y

LukasHirt:
## Description
- Refactor switch component
- Move styles directly to SFC
- Add unit tests
- Round corners
- Add label
- Use button

## Screenshots
image

## Open tasks:

- [x] unit tests


src/components/switch/OcSwitch.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <span class="oc-switch">
     <button
+      data-testid="oc-switch-btn"
```
kulmann, line 4: We have used `data-test-id` so far in this repo. You could start a discussion (in web discussions, because it applies there as well) if we want to go with `data-testid` instead and accept the fact that `data-test-some-other-prop` would have a different format then.

LukasHirt: https://github.com/owncloud/web/discussions/5367

src/components/switch/OcSwitch.vue
```diff
@@ -0,0 +1,154 @@
...
+  <span class="oc-switch">
+    <button
+      data-test-id="oc-switch-btn"
+      class="oc-switch-btn"
+      role="switch"
+      :aria-checked="state.toString()"
+      :aria-labelledby="labelId"
+      @click="toggle"
+    />
+    <span :id="labelId" v-text="label" />
```
kulmann, line 11: What do you think about using an actual `<label>` instead of a `<span>` as that's already the native html element that fulfills what you want to achieve? 🤔 To keep it consistent with our other form like components I'd prefer to have the label before the button (see e.g. `OcTextInput`).

LukasHirt: It's not used on purpose as label is not made for buttons

image

LukasHirt: Took a **great** inspiration here from Marcus https://github.com/marcus-herrmann/vue-switch-button/blob/master/src/components/SwitchButton.vue

kulmann (approved):
LGTM 🚀

PR #1440, opened 2021-06-30
added unit tests for OcAutocomplete component

kiranparajuli589:
## Description
added unit test for `OcAutocomplete` component
- only covered message part
- dropdown test will be added after UiKit removal

## Related Issue

- Part of #1366 

## How Has This Been Tested?

- 🤖 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/OcAutocomplete.spec.js
+++ b/src/components/OcAutocomplete.spec.js
@@ -1,0 +1,80 @@
+import OcAutocomplete from "./OcAutocomplete.vue"
+import { mount } from "@vue/test-utils"
+
+describe("OcAutocomplete", () => {
+  function getWrapperWithProps(props = {}) {
+    return mount(OcAutocomplete, {
+      propsData: {
+        ...props,
+        label: "Test Label",
+      },
+    })
+  }
+  const selectors = {
+    autocomplete: ".oc-autocomplete",
+    message: ".oc-autocomplete-message",
+    dropdown: ".oc-autocomplete-dropdown",
+  }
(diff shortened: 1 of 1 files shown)


src/components/OcAutocomplete.spec.js
```diff
@@ -0,0 +1,80 @@
...
+        expect(inputElement.attributes("class")).not.toContain("oc-autocomplete-warning")
+      })
+    })
+    describe("description message", () => {
+      it("should render message if prop 'descriptionMessage' is set", () => {
+        const wrapper = getWrapperWithProps({ descriptionMessage: "Help Text" })
+        const descriptionMessageElement = wrapper.find(selectors.message)
+        expect(descriptionMessageElement.exists()).toBeTruthy()
+        expect(descriptionMessageElement.isVisible()).toBeTruthy()
+        expect(descriptionMessageElement.get("span").text()).toBe("Help Text")
```
individual-it, line 54: can we do without the `span`? to not depend on that implementation detail?
```suggestion
        expect(descriptionMessageElement.text()).toBe("Help Text")
```

individual-it (changes requested):

src/components/OcAutocomplete.spec.js
kiranparajuli589: to assert text, we can actually.
but for class we've to get the span `el`

individual-it (approved):

PR #1452, opened 2021-06-30
added unit tests for OcNavbar component

kiranparajuli589:
## Description
added unit tests for `OcNavbar` component

## Related Issue

- Part of #1366 

## How Has This Been Tested?

- 👦🏼 
- 🤖 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/OcNavbar.spec.js
+++ b/src/components/OcNavbar.spec.js
@@ -1,0 +1,55 @@
+import OcNavbar from "./OcNavbar.vue"
+
+const { mount } = require("@vue/test-utils")
+
+describe("OcNavbar", () => {
+  function getWrapper(props = {}) {
+    return mount(OcNavbar, {
+      propsData: {
+        ...props,
+      },
+    })
+  }
+  const selectors = {
+    navContainer: ".uk-navbar-container",
+  }
+  describe("sticky", () => {
+    describe("when sticky if enabled", () => {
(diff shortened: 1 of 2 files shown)


src/components/OcNavbar.spec.js
```diff
@@ -0,0 +1,55 @@
...
+      const actualTagName = wrapper.find(selectors.navContainer).element.tagName
+      expect(actualTagName).toBe(tag.toUpperCase())
+      expect(wrapper.find(actualTagName)).toBeTruthy()
+    })
+  })
+  describe("slot", () => {
+    it("should render provided slot html", () => {
+      const wrapper = mount(OcNavbar, {
+        slots: {
+          default: '<p class="test">Test</p>',
```
individual-it, line 45: lets use strings that are more different than `test` and `Test` otherwise it might happen that if the content is taken into the class because of some bug and just switched to lower-case the test will not catch it
```suggestion
          default: '<p class="test class">Test content</p>',
```

individual-it (changes requested):

individual-it (approved):

kulmann (approved):

PR #1460, opened 2021-07-02
Switch from data-test-id to data-testid

pascalwengerter:
## Description
See https://github.com/owncloud/web/discussions/5367


kulmann (approved):
👍

PR #1521, opened 2021-07-22
Introduce search:input event to select, deprecate autocomplete and add OcRecipient

LukasHirt:
OcSearch can be used with prop `:multiple="true"` to achieve the same behaviour as our autocomplete and even better with including selected options directly in the select...

image

image


pascalwengerter:
License/CLA fails since @paulcod3 has changed his accounts and can't sign in his old account (from which the commit has been cherrypicked). Perhaps best to reset the changes locally and change authors from his old to his new account

changelog/unreleased/change-deprecate-autocomplete
```diff
@@ -0,0 +1,5 @@
+Change: deprecate OcAutocomplete
```
pascalwengerter, line 1: Deprecating isn't a change in itself, I'd vote in favor of having the deprecation as an enhancement in the changelog and then dropping all deprecated components in one of the upcoming major releases

package.json
```diff
@@ -159,5 +159,8 @@
   "engines": {
     "node": ">= 8.6.0",
     "npm": ">= 3.0.0"
+  },
+  "dependencies": {
```
pascalwengerter, line 163: Isn't this a `peerDependencies` like the others?

changelog/unreleased/change-deprecate-autocomplete
kulmann: I agree, dropping the component will be the change IMO

src/components/OcSelect.vue
```diff
@@ -187,4 +286,9 @@ export default {
 }
 </style>
 ```
+
+## Loading spinner
+```js
+<oc-select :options="['Apple', 'Bannana', 'Orange', 'Pear']" :multiple="true" :loading="true" />
```
pascalwengerter, line 292: This one breaks the dropdown-preview for me. Could you take a look? :)

src/components/OcAutocomplete.vue
```diff
@@ -683,9 +542,14 @@ export default {
         this.delayedResult = this.simpleItems
```
pascalwengerter, line 542: Delayed result example seems broken

pascalwengerter (changes requested):
Love a lot of what I'm seeing here, small questions about broken/outdated examples and we should be good to go!

src/components/OcAutocomplete.vue
LukasHirt: I threw away the changes as since we want to deprecate it anyway it doesn't make too much sense to introduce changes to it now...

pascalwengerter: Sure thing, then probably just remove the broken example? :)

src/components/OcSelect.vue
LukasHirt: Dropdown shouldn't be visible at all since it is in loading state (clarified in chat)

src/components/OcAutocomplete.vue
LukasHirt: It works now, was broken due to new changes.

kulmann (commented):
Running `yarn start` I get a lot of deprecation warnings like these `DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.`, while master looks fine. Can you confirm?

LukasHirt:
> I get a lot of deprecation warnings

Yep, some deps have been updated (no major version bump though). As it doesn't break yet anything, I would say let's take care of updating to new functions in a separate PR...

src/components/OcRecipient/OcRecipient.vue
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        if (typeof recipient.name !== "string") {
+          throw new Error("Recipient name is not a string")
+        }
+
+        if (recipient.name.length < 1) {
+          throw new Error("Recipient name is empty")
+        }
+
```
kulmann, line 66: Please add validation about if an `icon` key is present in the recipient, that it requires to have the sub keys `recipient.icon.name` and `recipient.icon.label`.

LukasHirt:
Some of them are btw coming from the UIkit... since we're not updating that one anymore due to some breaking changes it might be necessary to either not update Dart Sass to v2, change the decision and invest the effort to update UIkit even though we're going to remove it or finally move forward with sending it to not-in-our-project dependency heaven :shrug:

kulmann (approved):
Would be nice to have validation for the format of the `icon` inside the recipient. And the deprecation warnings of the updated dependencies need to be fixed in a separate PR. Rest looks good to me 👍 💪

pascalwengerter (approved):
LGTM 👥

PR #1525, opened 2021-07-22
Fix resource ID handling in OcTableFiles contextmenu

pascalwengerter:
## Description
Regex was wrong and actually unnecessary so it can be removed. Needs to be published for integration in `web`.

### DIFF
--- a/src/components/table/OcTableFiles.vue
+++ b/src/components/table/OcTableFiles.vue
@@ -82,20 +82,20 @@
         <!-- @slot Add quick actions directly next to the `showDetails` button in the actions column -->
         <slot name="quickActions" :resource="item" />
         <oc-button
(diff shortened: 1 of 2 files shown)


src/components/table/OcTableFiles.vue
```diff
@@ -82,20 +82,20 @@
         <!-- @slot Add quick actions directly next to the `showDetails` button in the actions column -->
         <slot name="quickActions" :resource="item" />
         <oc-button
-          :id="`quick-action-${item.id.replace(/=/, '')}`"
+          :id="`quick-action-${item.id.replace(/==/, '')}`"
```
kulmann, line 85: Isn't `/=*/` what you want to check?

pascalwengerter: True, fixed

kulmann (approved):

PR #1530, opened 2021-07-23
Fix a11y issues from release 7.1.3

kulmann:
DO NOT MERGE

This PR fixes the following a11y issues, branching off from version 7.1.3
<list following>


kulmann:
> @kulmann how do we handle this? Steve had originally signed the CLA and we have commits by him on master also, not sure why it fails now

CLA was originally signed at the time when the commits were made and we won't be able to cherry-pick any of this to master anyway. So IMO it just looks ugly on this PR now, which is only cosmetic.

PR #1533, opened 2021-07-23
Remove click handler from th for sorting

kulmann:
## Description
Remaining fixes for fixing the table sorting for a11y. Click handler for sorting was firing double (once from the `<th>` and once from the `<oc-button>`). Since the table heading is now in the button as well, we don't need the click handler on the `<th>` anymore.

Fixes part (a) of https://github.com/owncloud/web/issues/5380


paulcod3:
@kulmann CI failes at unittests

paulcod3:
LGTM :)

paulcod3 (approved):

PR #1534, opened 2021-07-23
a11y remove aria-label for oc-select

paulcod3:
## Description
See https://github.com/owncloud/web/issues/5394

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5394

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/OcSelect.vue
+++ b/src/components/OcSelect.vue
@@ -33,16 +33,6 @@ export default {
       default: null,
     },
   },
-
-  mounted() {
-    this.setComboBoxAriaLabel()
-  },
-  methods: {
-    setComboBoxAriaLabel() {
-      const comboBoxElement = this.$refs.select.$el.querySelector("div:first-child")
-      comboBoxElement.setAttribute("aria-label", this.$gettext("Search for option"))
-    },
-  },
 }
 </script>
 
--- a/changelog/unreleased/enhancement-oc-select-remove-aria-label
+++ b/changelog/unreleased/enhancement-oc-select-remove-aria-label
@@ -1,0 +1,7 @@
+Enhancement: OcSelect remove aria label
+
+Now the screenreader should read the actual label for the OcSelect rather than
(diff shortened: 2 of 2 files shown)


src/components/OcSelect.vue
```diff
@@ -33,16 +33,6 @@ export default {
...
     },
   },
-
-  mounted() {
-    this.setComboBoxAriaLabel()
-  },
-  methods: {
-    setComboBoxAriaLabel() {
-      const comboBoxElement = this.$refs.select.$el.querySelector("div:first-child")
-      comboBoxElement.setAttribute("aria-label", this.$gettext("Search for option"))
```
fschade, line 43: instead of removing it it could be a better approach to use it as a fallback label if no label is provided

paulcod3: hm i looked up all places where oc-select is used in web and they all set a label (with aria-label) outside of the oc-select, i think it makes more sence to leave the aria-labeling to the implementor since oc-select doesn't include a label.

pascalwengerter: @kulmann @paulcod3 if we plan to merge this it'd be a breaking change and I would suggest cherry-picking it and bringing it inside the `v9.0.0` release this week, too 🍒

pascalwengerter:
@kulmann what's the way to go here? We could either drop the label as Paul suggested or "bring it in" to satisfy owncloud/web#8857. From what I know there's only two occurances in `web` where we're using it and we could easily re-use the respective computed properties to pass them to the `OcSelect`

kulmann:
I'd like to work towards https://github.com/owncloud/web/issues/8857 to have a consistent component API for form-related elements.

pascalwengerter:
Okay so closing this in order to provide a solution that works towards owncloud/web#8857

Issue #538, opened 2019-11-12
Remove uikit


PVince81:
- [ ] Replace uikit classes with own ones where needed. For example: "uk-width-small" => "oc-width-small"
- [ ] Redefine uikit styles for now in "oc-" namespace (which can be adjusted later by a designer)
- [ ] Remove usage of uikit across Phoenix and make sure we have replacement "oc-" classes to use there if needed

@DeepDiver1975 @LukasHirt

DeepDiver1975:
furthermore I'd welcome to use tags to define things like layout instead of having to append many classes. See vuetify ....

PR #1545, opened 2021-07-30
Drop deprecated components

pascalwengerter:
## Description
Still to be discussed if we should give a warning in the next major release and only drop them in the major release that follows after, but we should have more focus and cleanup the design system


pascalwengerter:
This also works heavily towards dropping uikit (at least the JS side of it), works towards #538

kulmann (approved):
👍

PR #1559, opened 2021-08-03
Web-853: Group files and folder when sorting by name

paulcod3:
## Description
See https://github.com/owncloud/web/issues/5613

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests updated
- [ ] Acceptance tests added

### DIFF
--- a/src/components/table/mixins/sort.js
+++ b/src/components/table/mixins/sort.js
@@ -19,43 +19,55 @@ export default {
       if (!this.isSortable || !this.sortBy) {
         return
       }
-
-      return [...this.data].sort((a, b) => {
-        let aValue = a[this.sortBy]
-        let bValue = b[this.sortBy]
-        const modifier = this.sortDir === SORT_DIRECTION_ASC ? 1 : -1
-
-        const { sortable } = this.fields.find(f => f.name === this.sortBy)
-
-        if (sortable) {
-          if (typeof sortable === "string") {
-            const genArrComp = vals => {
-              return vals.map(val => val[sortable]).join("")
-            }
-
-            aValue = genArrComp(aValue)
-            bValue = genArrComp(bValue)
-          } else if (typeof sortable === "function") {
-            aValue = sortable(aValue)
-            bValue = sortable(bValue)
-          }
+      if (this.sortBy === "name") {
+        const folders = [...this.data.filter(i => i.type === "folder")].sort((a, b) =>
+          this.sortData(a, b)
+        )
+        const files = [...this.data.filter(i => i.type === "file")].sort((a, b) =>
+          this.sortData(a, b)
+        )
+        if (this.sortDir === SORT_DIRECTION_ASC) {
+          return folders.concat(files)
         }
-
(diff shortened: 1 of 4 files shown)


changelog/unreleased/enhancement-sort-table-files-folders
```diff
@@ -0,0 +1,6 @@
+Enhancement: sort table files and folders
```
pascalwengerter, line 1: The changelog item name / title suggests you're introducing sorting to the table

I'd prefer making more explicit what gets added here, e.g. "Group files and folder when sorting by name"

paulcod3: yes that sounds better thx ^^

pascalwengerter:
@paulcod3 the `Fixes #issue-nr` in the PR description automatically closes the issue in the `web` repo once this PR gets merged in the design system. However this is not the case since the ODS needs a release and a version bump merged in the `web` master before the issue can be considered closed. I usually document this as "Works towards #issue-nr" so it still gets linked within GH but not auto-closed 😉

paulcod3:
@pascalwengerter ah thanks you are right that was the problem last time ^^

src/components/table/mixins/sort.js
```diff
@@ -19,43 +19,54 @@ export default {
       if (!this.isSortable || !this.sortBy) {
         return
       }
+      const hasFilesOrFolders = this.data.some(i => i.type === "folder" || i.type === "file")
```
pascalwengerter, line 22: Can't really dig into the code right now but since this is used only in `OcTableFiles` which will _always_ contain of either *"folder"s* or *"file"s* I don't see a case where this statement won't be true.

Would `hasFilesAndFolders` with a logical `&&` instead of the `||` make more sense maybe?

paulcod3: yeah you are right, I will check if its only used in `OcTableFiles` and if it is I'll remove this check.

paulcod3: ok i looked it up and you are right, its only use is with OcTableFiles, so I removed this check.

src/components/table/mixins/sort.js
```diff
@@ -19,43 +19,53 @@ export default {
       if (!this.isSortable || !this.sortBy) {
         return
       }
+      if (this.sortBy === "name") {
+        let folders = this.data.filter(i => i.type === "folder")
```
kulmann, line 23: I'd like to see filtering and sorting chained and (because it's not modified after that) stored in a `const`:
`const folders = this.data.filter(i => i.type === "folder").sort((a, b) => this.sortData(a, b))`

Same for files.

kulmann (changes requested):

src/components/table/mixins/sort.js
paulcod3: yes good improvement

kulmann (approved):
LGTM 👍

Issue #1560, opened 2021-08-03
Drag & Drop doesn't select files correctly


paulcod3:
### Expected behaviour
When dragging a file in the octablefiles it should work with one or multiple files.

### Actual behaviour
Due to refactoring this functionality got broken.

PR #1562, opened 2021-08-03
Bugfix: Drag & Drop doesn't select files correctly

paulcod3:
## Description
See #1560 

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/bugfix-selected-files-drag-drop
+++ b/changelog/unreleased/bugfix-selected-files-drag-drop
@@ -1,6 +1,6 @@
 Bugfix: Drag & Drop doesn't select files correctly
 
-Due to refactoring the selection of the files when dragging doesn't work right.
+Due to refactoring there was a case where selecting files when dragging didn't work right. This has been addressed and fixed now.
 
 https://github.com/owncloud/owncloud-design-system/issues/1560
 https://github.com/owncloud/owncloud-design-system/pull/1562
--- a/changelog/unreleased/bugfix-selected-files-drag-drop
+++ b/changelog/unreleased/bugfix-selected-files-drag-drop
@@ -3,4 +3,4 @@ Bugfix: Drag & Drop doesn't select files correctly
 Due to refactoring the selection of the files when dragging doesn't work right.
 
 https://github.com/owncloud/owncloud-design-system/issues/1560
-https://github.com/owncloud/owncloud-design-system/pull/1561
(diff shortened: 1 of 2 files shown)


changelog/unreleased/bugfix-selected-files-drag-drop
```diff
@@ -0,0 +1,6 @@
+Bugfix: Drag & Drop doesn't select files correctly
+
+Due to refactoring the selection of the files when dragging doesn't work right.
```
pascalwengerter, line 3:
```suggestion
Due to refactoring there was a case where selecting files when dragging didn't work right. This has been addressed and fixed now.
```

pascalwengerter (commented):
Changes LGTM, wording in the changelog item should give more context 😇 sorry for being naggy!

paulcod3:
@pascalwengerter yes you are right, was just trying to fix it as fast as possible

src/components/table/OcTableFiles.vue
```diff
@@ -382,7 +382,7 @@ export default {
   methods: {
     fileDragged(file) {
       const selectedResourceInResources = this.selectedResources.some(e => e.id === file.id)
-      if (selectedResourceInResources) {
+      if (!selectedResourceInResources) {
```
pascalwengerter, line 385: Just gave it a try (by building it locally and linking it to https://github.com/owncloud/web/pull/5588) and I don't think that's the right way to go, at least for single file drag&drop: As soon as you start dragging your single file of choice, the `Move` and `Copy` buttons appear above the table (since, once you start dragging, you now have a selected resource) causing the UI to become "jumpy"

pascalwengerter (changes requested):

src/components/table/OcTableFiles.vue
paulcod3: Could we discuss a solution if you have time? I'm not really sure what we can do about this since we need the selection in order to move files.

pascalwengerter: Figured out that this "jumpy" behaviour will get fixed by moving the bulk actions into the right sidebar so good for now

pascalwengerter (approved):
I'd love to have tests for this but should be good for now

PR #1568, opened 2021-08-05
Web-648: OcTableFiles enable multiple highlighting using selection

paulcod3:
## Description
See https://github.com/owncloud/web/issues/5164

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/change-enable-multiple-highlighting
+++ b/changelog/unreleased/change-enable-multiple-highlighting
@@ -1,0 +1,7 @@
+Change: Enable multiple highlighting for OcTableFiles
+
+We changed the highlighting in a way that now every selected file
+is highlighted automatically. The `highlighted` prop has been removed as it's not used anymore. 
+
+https://github.com/owncloud/web/issues/5164
+https://github.com/owncloud/owncloud-design-system/pull/1568
--- a/src/components/table/OcTableFiles.vue
+++ b/src/components/table/OcTableFiles.vue
@@ -7,7 +7,7 @@
     :sticky="true"
     :header-position="headerPosition"
     :drag-drop="dragDrop"
-    @highlight="showDetails"
+    @highlight="fileClicked"
     @rowMounted="rowMounted"
     @contextmenuClicked="showContextMenu"
     @fileDropped="fileDropped"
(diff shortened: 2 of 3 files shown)


changelog/unreleased/change-enable-multiple-highlighting
```diff
@@ -0,0 +1,7 @@
+Change: Enable multiple highlighting for OcTableFiles
+
+We changed the highlighting in a way that now every selected file
+is highlighted.
```
kulmann, line 4: Please add description about why this is a breaking change, e.g. the following:
```suggestion
is highlighted automatically. The `highlighted` prop has been removed as it's not used anymore. 
```

src/components/table/OcTableFiles.vue
```diff
@@ -379,6 +371,12 @@ export default {
     areAllResourcesSelected() {
       return this.selection.length === this.resources.length
     },
+
+    selectedIds() {
+      let ids = []
```
kulmann, line 376: Computed prop can be solved as a one liner:
`return this.selectedResources.map(r => r.id)`

kulmann (changes requested):
Cool, didn't expect the changes to be so minimal. Awesome! Added some ideas for improvement.

src/components/table/OcTableFiles.vue
paulcod3: yes that looks better

kulmann:
As discussed in a call:
- the `highlight` event from `OcTable` needs to be used to update the selection model. don't emit `showDetails` anymore.
- in oC Web we'll watch the selection model to decide whether or not the right sidebar needs to be shown
- the `show details` button needs a new implementation: stop the propagation of the click event and add a new method for it, which resets the selection model to the resource id of the clicked row/button (in oC Web this results in a right sidebar for the clicked row)

paulcod3:
I would only change the unittests so they work for now since the highlighting logic will changed again soon.

kulmann (approved):
LGTM 👍

Issue #1503, opened 2021-07-14
Label option for ocSelect missing


elizavetaRa:
### Expected behaviour
Option for label for ocSelect

### Actual behaviour
Without this option, trying to implement label left from ocSelect with wrapper-style:
```
  display: flex;
  flex-direction: row;
  align-items: center;
```
has unexpected behavior for ocSelect, related issue https://github.com/owncloud/owncloud-design-system/issues/1502

### Screenshots (if appropriate):
![Files shared with me - ownCloud (1)](https://user-images.githubusercontent.com/7430156/125579739-6ab5218a-c75b-40b8-b1ff-dbd371bfbe5d.gif)

PR #1570, opened 2021-08-05
Add label to OcSelect

pascalwengerter:
## Description
See changelog item

### DIFF
--- a/changelog/unreleased/enhancement-ocselect-label
+++ b/changelog/unreleased/enhancement-ocselect-label
@@ -1,0 +1,5 @@
+Enhancement: Label for OcSelect
+
+We've added a configurable `<label>` for the OcSelect label.
+
+https://github.com/owncloud/web/issues/1570
--- a/changelog/unreleased/enhancement-ocselect-label
+++ b/changelog/unreleased/enhancement-ocselect-label
@@ -1,5 +1,0 @@
-Enhancement: Label for OcSelect
-
-We've added a configurable `<label>` for the OcSelect label.
-
-https://github.com/owncloud/web/issues/1570
--- a/src/components/OcSelect.vue
+++ b/src/components/OcSelect.vue
@@ -1,11 +1,12 @@
 <template>
   <div>
-    <label :for="this.$attrs['input-id']" v-text="setLabel" />
+    <label :for="id" v-text="label" />
     <vue-select
       ref="select"
+      :input-id="id"
       :value="model"
       class="oc-select"
-      v-bind="setCustomizedAttrs()"
+      v-bind="additionalAttributes"
       v-on="$listeners"
     >
       <template v-for="(index, name) in $scopedSlots" #[name]="data">
@@ -17,6 +18,7 @@
 </template>
 
 <script>
+import uniqueId from "../utils/uniqueId"
 import VueSelect from "vue-select"
 import "vue-select/dist/vue-select.css"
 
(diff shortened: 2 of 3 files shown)


pascalwengerter:
Cherrypicking this commit to `master` will close owncloud/owncloud-design-system#1503

src/components/OcSelect.vue
```diff
@@ -32,6 +35,11 @@ export default {
       required: false,
       default: null,
     },
+    labelText: {
```
kulmann, line 38: Would it make sense to bind `$attrs` without the `label` prop to the vue-select? So that we can have a dedicated `label` prop on the OcSelect? I kind of dislike using `labelText` as prop name here, while we use `label` on all other form related elements.

pascalwengerter:
@kulmann ready for re-review. Feels a bit hacky and prevents us from using the [vue-select label prop](https://vue-select.org/api/props.html#label) but gets the job done

src/components/OcSelect.vue
```diff
@@ -33,6 +42,11 @@ export default {
       default: null,
     },
   },
+  computed: {
+    setLabel() {
```
paulcod3, line 46: i find this somehow misleading, sounds like a method. Maybe selectLabel?

pascalwengerter: How about `inputLabel`? @kulmann any suggestions/opinions?

kulmann: yeah, `inputLabel` is fine

src/components/OcSelect.vue
```diff
@@ -42,6 +56,11 @@ export default {
       const comboBoxElement = this.$refs.select.$el.querySelector("div:first-child")
       comboBoxElement.setAttribute("aria-label", this.$gettext("Search for option"))
     },
+    setCustomizedAttrs() {
```
kulmann, line 59: This is not setting anything. It should be a computed prop (and have a different name, like `customizedAttrs` or something like that)

kulmann: Also, you should copy the `this.$attrs` first before modifying.

changelog/unreleased/enhancement-ocselect-label
```diff
@@ -0,0 +1,5 @@
+Enhancement: Label for OcSelect
+
+We've added a configurable `<label>` for the OcSelect label.
```
dschmidt, line 3: Tbh this sounds a bit strange.

You've just added a label `for OcSelect`, no?

changelog/unreleased/enhancement-ocselect-label
```diff
@@ -0,0 +1,5 @@
...
+
+https://github.com/owncloud/web/issues/1570
```
dschmidt, line 5: Should be #1503 I think

src/components/OcSelect.vue
```diff
@@ -147,16 +166,18 @@ export default {
 ```
 
 ### Using slots to display complex options
-Sometimes we need to display more complex options. This can include e.g. an option with a title and a description. To still display all those values exactly as we want to, we need to use scoped slot called `option`.
-We can then retrieve all the values that we want to display from the slots parametres. It is important to specify the `label` prop on the `oc-select` component which will specify which key should be used as the option label.
+Sometimes we need to display more complex options. This can include e.g. an option with a title and a description. 
+To still display all those values exactly as we want to, we need to use scoped slot called `option`.
+We can then retrieve all the values that we want to display from the slots parametres. 
+It is important to use a `label` key inside the passed object since `oc-select` will look for it.
```
dschmidt, line 172: This sounds rather complicated. Why don't we introduce a `option-label` property that will be passed as `label` to vue-select?

dschmidt (changes requested):
IMHO this is a breaking change (I'm using the `label` property already) and should be marked as such in the changelog.

I would prefer having an `option-label` prop instead of having to use the `option` slot but that's obviously debatable.

dschmidt:
In general: yay!
With this we can replace one more ❌ with ✔️ in https://github.com/owncloud/web/issues/8857  :-D
Much appreciated if you could do that once this gets merged.

src/components/OcSelect.vue
pascalwengerter: Sounds like a great idea actually, let me take a look 🕵🏽

dschmidt: Or get it from `$attrs` fwiw

src/components/OcSelect.vue
dschmidt: Full ack! Streamline all the things 🥳

changelog/unreleased/enhancement-ocselect-label
dschmidt: Ah, nope - https://github.com/owncloud/owncloud-design-system/pull/1570 hehe

src/components/OcSelect.vue
```diff
@@ -36,31 +36,50 @@ export default {
...
       required: false,
       default: null,
     },
-  },
-  computed: {
-    setLabel() {
-      return this.$attrs["label"]
+    /**
+     * Label of the select component
+     * ATTENTION: this shadows the vue-select option `label`. If you need access to that use `optionLabel`.
```
dschmidt, line 49: It shadows the vue select "prop"

src/components/OcSelect.vue
```diff
@@ -1,10 +1,19 @@
 <template>
-  <vue-select ref="select" :value="model" class="oc-select" v-bind="$attrs" v-on="$listeners">
-    <template v-for="(index, name) in $scopedSlots" #[name]="data">
-      <slot :name="name" v-bind="data"></slot>
-    </template>
-    <div slot="no-options" v-translate>No options available.</div>
-  </vue-select>
+  <div>
+    <label :for="this.$attrs['input-id']" v-text="label" />
```
dschmidt, line 3: Where is this supposed to come from? How is this wired up with vue-select?

dschmidt: Ah I assume vue-select has this prop? In other places we seem to use an id prop, maybe we should stick to that

dschmidt: https://vue-select.org/api/props.html#inputid

dschmidt: https://github.com/owncloud/owncloud-design-system/blob/17e7e6fd968a536f56f91db8d30a0317535583b3/src/components/OcTextInput.vue#L70

dschmidt:
CI fails because of unit tests, looks completely unrelated to me. `OcSelect` does not have unit tests after all :trollface:

pascalwengerter:
Squash-merging this into the a11y-updates branch now as it has the same changeset as owncloud/owncloud-design-system#1633 (@dschmidt your review was implemented, thanks for the constructive feedback and collaboration!)

PR #1574, opened 2021-08-06
Bugfix: OcTableFiles selection model

paulcod3:
## Description
We used a local prop to hold the state of the selection model of octablefiles which turned out to be wrong.
Should be fixed with this PR.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/table/OcTableFiles.vue
+++ b/src/components/table/OcTableFiles.vue
@@ -261,7 +261,6 @@ export default {
       constants: {
         EVENT_TROW_MOUNTED,
       },
-      selectedResources: [],
     }
   },
   computed: {
@@ -373,7 +372,7 @@ export default {
     },
 
     selectedIds() {
-      return this.selectedResources.map(r => r.id)
+      return this.selection.map(r => r.id)
     },
   },
   methods: {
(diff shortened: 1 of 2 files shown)


src/components/table/OcTableFiles.vue
```diff
@@ -630,16 +629,6 @@ export default {
             type: "folder",
             size: "5324435",
             mdate: "Mon, 11 Jan 2021 14:34:04 GMT"
-          },
```
pascalwengerter, line 633: Why delete this example?

pascalwengerter (changes requested):
1. Missing changelog item (minor thing)
2. Question as commented
3. Works great in general, but was able to cause a weird issue:

- Select multiple files via checkbox
- Open Details via Contextmenu
- Unselect all items via the "Clear selection/X" button
- OcTableFiles only shows `name` and `quickactions` but not `size` and `last modified` anymore 🤷🏽

src/components/table/OcTableFiles.vue
```diff
@@ -384,11 +383,12 @@ export default {
       this.$emit(EVENT_FILE_DROPPED, fileId)
     },
     addSelectedResource(file) {
-      const selectedResourceInResources = this.selectedResources.some(e => e.id === file.id)
+      const selectedResourceInResources = this.selection.some(e => e.id === file.id)
```
kulmann, line 386: Seeing this again now, what do you think about `isSelected` als name for the const?

src/components/table/OcTableFiles.vue
kulmann: If this was deleted in the gist I sent to you @paulcod3 it was not intentional. Please restore. ;-)

kulmann (changes requested):
As discussed, add the PR URL to the already existing `enhancement` changelog item. That's sufficient, as this code with the highlighting-changes was never released. IMO dedicated bugfix changelogs make sense after something was released before.

Bug that Pascal mentions is weird and needs to be double checked. I will also take the time and run this in your open web PR soonish (on Monday at the latest).

paulcod3:
Ok found the solution to the following two issues:
1. Sidebar wasn't going to open when selecting a file
2. When opening the sidebar and deselecting all table columns went missing

As a solution we need to refactor some web logic since the "highlightedFile" getter isn't valid anymore.
@kulmann I would discuss the solution when you have time since it affects existing logic

kulmann (approved):
LGTM 👍

PR #1575, opened 2021-08-08
feat: drop custom target and manual mode

LukasHirt:


src/components/OcDrop.vue
```diff
@@ -91,14 +101,18 @@ export default {
   },
   mounted() {
     destroy(this.tippy)
-    const to = this.toggle ? document.querySelector(this.toggle) : this.$el.previousElementSibling
+    let to = this.target
```
kulmann, line 104: Can be a `const`

src/components/OcDrop.vue
```diff
@@ -113,14 +127,40 @@ export default {
...
         this.tippy.hide()
       }
     },
+
+    /**
+     * Programatically show the drop
+     *
+     * @public
+     */
+    $_ocDrop_show(duration) {
```
kulmann, line 152: Since this is public `show` as method name would be fine, right?

src/components/OcDrop.vue
```diff
@@ -113,14 +127,40 @@ export default {
...
+      this.tippy.show(duration)
+    },
+
+    /**
+     * Programatically hide the drop
+     *
+     * @public
+     */
+    $_ocDrop_hide(duration) {
```
kulmann, line 161: Since this is public `hide` as method name would be fine, right?

src/components/OcDrop.vue
LukasHirt: I was thinking about this in a scope of a mixin where we also did this pattern. In the Web PR I am accessing it via ref but TBH I am not sure if the method can in any way leak out of the component

kulmann: For a mixin the namespacing makes sense, because you can't see where the method comes from when calling it. For calling it via a ref, you know where it comes from, so I personally don't need the namespacing.

Not sure either if the method can leak out in any way, but it is intentionally public, so I think we don't need to think about that, do we? `Intentionally public` evaluates to `drop the '$_'` in my mind. 🤷

kulmann (approved):
Awesome 👌🏼

PR #1576, opened 2021-08-09
release: 9.0.0

LukasHirt:


kulmann (changes requested):
Please copy the PR link from the bugfix into the `enhancement-table-drag-drop` changelog item. Since the feature is released for the first time it doesn't make sense to have bugfixes about it in the same release. And then delete the bugfix changelog item.

kulmann (approved):
👍

PR #1593, opened 2021-08-17
fix(ocDrop bugfix): update buggy drop example

fschade:
drop example used undefined $_ocDrop_show method, update to use only
show


pascalwengerter (approved):

pascalwengerter:
Nice catch!

PR #1609, opened 2021-08-23
Pin sass version to 1.32.12 again and ignore in dependabot

kulmann:
Pin sass version to 1.32.12 again and ignore in dependabot config.

Also cleans up the ignored dependencies in the dependabot config.


pascalwengerter (approved):
Changes LGTM, thanks for taking care! Also closes my revert-PR

PR #1625, opened 2021-09-02
Reset row background on dragDrop event in OcTable

pascalwengerter:
## Description
Thanks @paulcod3 for the hint, wasn't visible in the ODS but got discovered once the `EVENT_FILE_DROPPED` actually did something inside `web`

### DIFF
--- a/src/components/table/OcTable.vue
+++ b/src/components/table/OcTable.vue
@@ -46,7 +46,7 @@
         "
         @hook:mounted="$emit(constants.EVENT_TROW_MOUNTED, item, $refs[`row-${trIndex}`][0])"
         @dragstart.native.stop="dragStart(item)"
-        @drop.native.stop="dropRowEvent"
+        @drop.native.stop="dropRowEvent(item.id, $event)"
         @dragenter.native.prevent.stop="dropRowStyling(item.id, false, $event)"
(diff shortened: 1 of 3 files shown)


src/components/table/OcTable.vue
```diff
@@ -249,15 +249,16 @@ export default {
...
       if (!this.dragDrop) return
       const dropTarget = event.target
       const dropTargetTr = dropTarget.closest("tr")
       const dropFileId = dropTargetTr.dataset.fileId
+      this.dropRowStyling(id, true, event)
       this.$emit(EVENT_FILE_DROPPED, dropFileId)
     },
     dropRowStyling(id, leaving, event) {
-      if (event.currentTarget.contains(event.relatedTarget)) {
+      if (event.currentTarget !== null && event.currentTarget.contains(event.relatedTarget)) {
```
kulmann, line 261: Since we're using node >= 14 you could do this with optional chaining:
```suggestion
      if (event.currentTarget?.contains(event.relatedTarget)) {
```

kulmann (changes requested):

src/components/table/OcTable.vue
pascalwengerter: Good hint, thanks!

kulmann (approved):

PR #1626, opened 2021-09-02
refactor(pagination): use route query object instead of param

fschade:
instead of having current pagination page as param refactor it to
use the query route object

BREAKING CHANGE: this could break in some cases where the implementing
logic relies on route page param, to fix switch to use route query
object


changelog/unreleased/enhancement-pagination-page-param
```diff
@@ -0,0 +1,5 @@
+Enhancement: Use route query to store active page
```
kulmann, line 1: Breaking change, isn't it?

kulmann (approved):

PR #1627, opened 2021-09-02
Prepare 10.0.0 release

pascalwengerter:
## Description
This PR contains
- release preparation for v10.0.0 (release already done)
- drone config changes so that we don't run steps repeatedly in different pipelines of the same CI run. With this PR being merged we'll only have one pipeline which has a dynamic set of steps, depending on whether this is a release, a PR or a merge commit to master.


.drone.star
```diff
@@ -26,192 +22,161 @@ def testing(ctx):
...
-                    'yarn build:docs',
-                ],
-                'depends_on': ['unit tests', 'eslint', 'stylelint']
+                'yarn build:system',
+            ],
+            'depends_on': ['unit tests']
+        },
+    ]
+
+    release_steps = [
```
kulmann, line 91: Release steps are only used on tag. IMO better to move construction of this array into the `if` statement body starting in line 153 (set the name first, then add all these additional release steps to the final_steps array).

kulmann (commented):
In general 👍 for the release stuff.

I don't have the drone knowledge to review the drone changes. Especially if reacting on the three different refs is used properly. 🤷 Will add more reviewers.

phil-davis (approved):
Looks OK to me. @pascalwengerter you and @kulmann can decide if you want to take his suggestion in the comment. There is not a performance issue in running this amount of starlark, so no bog deal if an array is created and then ignored. Do whatever.

pascalwengerter:
> Looks OK to me. @pascalwengerter you and @kulmann can decide if you want to take his suggestion in the comment. There is not a performance issue in running this amount of starlark, so no bog deal if an array is created and then ignored. Do whatever.

Thanks for the review, from my POV it reduced the cognitive load when reasoning about the CI structure but not sure if that's a shared sentiment. Merging it now anyways to get this into `master` (and update changelog & docs)

PR #1633, opened 2021-09-04
Add label to OcSelect  (port to master)

dschmidt:
## Description
Ported #1570 to master.

## Related Issue

- Fixes #1503

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added/updated

## Open tasks:

- [ ] ...

PR #1634, opened 2021-09-04
OcSelect: add support for showing default values

dschmidt:
(depends on https://github.com/owncloud/owncloud-design-system/pull/1633)

## Description
I've added an optional`defaultValue` prop to `OcSelect`.
The value provided there is selected when `value === null`. When the default value is shown the clear button is hidden because you obviously can't clear the default.
It's possible to select the default value which then fires the `input` event, i.e. a user can override a default value by explicitly setting it.

I'm not really happy with the usability of this as it is right now for multiple reasons:
1) clicking an `x` and then seeing some value is confusing
2) clicking an `x` and then still seeing the same value is even more confusing
    (when the default was explicitly selected by the user)
3) it's not obvious what the default value is as soon as something is selected, c.f. placeholder for `OcTextInput` and `OcTextarea`

I figured it might be a bit better if we would use a different icon for "reverting to default", e.g. something like the replay icon provided by ODS. Unfortunately vue-select does not seem to provide api to replace the clear button: https://github.com/sagalbot/vue-select/blob/4230bba9349c5e3ba33a4e77a566fe3ff7d2e1c5/src/components/Select.vue#L58

If we decide to somehow display the default outside of the actual select component we need to make sure we respect `label` and possibly other relevant props.

We need to decide if we want to settle on `clearButtonEnabled` or `clearable` (...or whatever) as name for this kind of property.
`vue-select` uses `clearable`, `OcTextInput` uses `clearButtonEnabled`. I'm fine with whatever, although I personally like `clearable` a bit better I chose `clearButtonEnabled` for not having to introduce a breaking change in `OcTextInput`... 

## Related Issue

- Fixes <issue_link>

## Motivation and Context
I want to show a default value in ownBrander 2.0 if a user hasn't selected a value yet.

It's possible to handle this outside of the OcSelect component, but that's rather cumbersome and dealing with the vue-select events seem a bit brittle, which is why I would prefer to have this solved for once in ODS. With the additional benefit that my code wouldn't depend on vue-select and vue-select could be replaced better (if neccessary one day).

Moreover, this needs to be thought through UX wise and having a solution in ODS would avoid different approaches in different places.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

Showing the default:
Screenshot_20210905_004208
Default selected explicitly:
Screenshot_20210905_004257

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


src/components/OcSelect.vue
```diff
@@ -195,6 +301,28 @@ prevent clearing the selected value by hitting `delete`.
...
+### Providing a default value
+We can provide a `defaultValue` to `oc-select` that is preselected when `value` is `null`.
+When the `defaultValue` is displayed the clear button is automatically disabled.
+
+```js
+<template>
+  <div class="oc-docs-width-medium">
+    <oc-select v-model="selected" :options="['Apple', 'Bannana', 'Orange', 'Pear']" default-value="Orange" />
+    <p>
+      Selected: {{ selected === null ? "null" : selected }}
```
dschmidt, line 313:
```suggestion
      Selected: {{ selected || "null }}}"
```

src/components/OcSelect.vue
```diff
@@ -1,27 +1,36 @@
...
-    <template v-for="(index, name) in $scopedSlots" #[name]="data">
-      <slot v-if="name !== 'search'" :name="name" v-bind="data" />
-    </template>
-    <div slot="no-options" v-translate>No options available.</div>
-    <template #spinner="{ loading }">
-      <oc-spinner v-if="loading" />
-    </template>
-  </vue-select>
+  <div>
+    <label :for="id" v-text="label" />
```
pascalwengerter, line 3:
```suggestion
    <label :for="id" class="oc-label" v-text="label" />
```

Just checked how the update journey in `web` would look like and we seem to give all labels this class, which doesn't style anything initially but sometimes is used to target the label for styling

dschmidt: Makes sense but possibly belongs into the original pr

pascalwengerter: True, but that one is merged already (and I somehow ended here without understanding this _is not the original_ ;D my bad, could you include it either way?

Issue #1350, opened 2021-06-02
Streamline label and description-message props for all form components


dschmidt:
Component | label | description-message | Notes
------------ | ------------- | ------------- | -------------
OcTextInput | ✔️  | ✔️  | 
OcCheckbox | ✔️ | ❌ |
OcSelect | ❌ | ❌ | 💥  vue-select already has a prop called label<br /> possibly related https://github.com/owncloud/owncloud-design-system/issues/1289
OcTextArea | ✔️  | ✔️ |
OcAutocomplete | ✔️  | ✔️ | 
OcDatepicker | ✔️  | ✔️ | 
OcSwitch | ✔️  | ❌ | label is only used for aria-label but not visible

While at it, it could be nice to have a common css class for all labels/description so it's easier to style all at the same time.

dschmidt:
Added `defaultValue` to the table.

dschmidt:
Additionally I would argue it would make sense to make the clear action emit `input` and `change` events with `null` values.
That way clearing can be differentiated from entering empty values.

The clear button on `OcTextInput` currently emits an empty string via `input` and doesn't emit `change` at all.

PR #1635, opened 2021-09-05
OcCheckbox: add support for showing default values and clear buttons

dschmidt:
## Description
This PR adds support for showing default values and a clear button to `OcCheckbox`.

It's not beautiful (visually) by any means and the same UX problems as for #1634 apply here too.

Needs UX concept by @kulmann and @tbsbdr 

## Related Issue
- #1634
- Fixes 3 :x: from  #1350
    (`@change`, `clearButtonEnabled` and `defaultValue`)

## Motivation and Context
See #1634 

Nothing selected by user
Screenshot_20210905_124232
Clicked once to switch value:
Screenshot_20210905_124309
Clicked twice to explicitly set value:
Screenshot_20210905_124336

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/OcCheckbox.vue
+++ b/src/components/OcCheckbox.vue
@@ -11,6 +11,15 @@
       :disabled="disabled"
     />
     <label :for="id" :class="labelClasses" v-text="label" />
+    <oc-button
+      v-if="showClearButton"
+      :aria-label="clearButtonAccessibleLabelValue"
+      appearance="raw"
+      @click="onClear"
+    >
+      <oc-icon name="close" size="small" variation="passive" />
+    </oc-button>
+
   </span>
 </template>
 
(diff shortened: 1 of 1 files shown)


src/components/OcCheckbox.vue
```diff
@@ -91,36 +128,38 @@ export default {
...
       return {
         "oc-invisible-sr": this.hideLabel,
         "oc-cursor-pointer": !this.disabled,
       }
     },
-  },
-  created() {
-    this.setChecked(this.model)
+    showClearButton() {
+      return this.clearButtonEnabled && this.value !== null
```
dschmidt, line 153: `this.disabled` needs to be taken into account.

tbsbdr:
Hey, this is admittedly a hard one from a ux perspective, especially the reset-x for a checkbox might feel odd.

**Requirements** as I understood them:
1. **Reset Option:** Provide an option to reset to the default
2. **Show if Default:** Show at a glance if the default is currently selected or not

I'm not aware of a common pattern, that meets both requirements for checkboxes. But maybe this ideas might help:

### Use OcSelect
Just don't use checkboxes at all for this usecase. Utilise your OcSelect instead like this:
image

### Use text-button
As it might not be 100% expectable what the reset-x will do, a textbutton might make the action more predictable.
image


**But:** Ideas 3-5 don't meet the 2nd requirement "Show if Default".
In VSCode Settings, they tried to highlight changed options with the blue vertical line, which is okay but not the best understandable solution, as you need to learn what the color means.

**Highlight changes with color:**

PR #1636, opened 2021-09-05
OcTextInput: add defaultValue prop and make clear button null the value

dschmidt:
## Description
1) Fixed the clear button being visible even when element is disabled
2) Made clear button emit `null` instead of empty string
3) Made input/change event handling a bit more consistent
4) Added `defaultValue` prop that is for now passed as `placeholder` to the input field. 

Needs UX concept by @kulmann and @tbsbdr

## Related Issue

- Fixes one :x: from #1350 (`defaultValue`) and makes it more consistent with other components that also emit `null` on clear (which is not part of that table)

## Motivation and Context
Same as #1634 and #1635 

Instead of using native `placeholder` prop on the `input` element, I would propose to introduce the `defaultValue` property which for now uses `placeholder` but keeps that as implementation detail. That way I can already use the functionality in ownBrander 2.0 and will automatically benefit from a better implementation at some point in the future.

Only visual change: it's now possible to have an empty input field with clear button.
That is useful when a user overrides the default value with an empty string.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
     we don't emit empty strings anymore on clear, but afaict `clearButtonEnabled` is used nowhere, so this shouldn't break anything in ownCloud
- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/enhancement-octextinput-props
+++ b/changelog/unreleased/enhancement-octextinput-props
@@ -1,0 +1,10 @@
+Enhancement: Streamline OcTextInput
+
+We have updated the OcTextInput component to be in line with other form components in the design system.
+
+- Fixed the clear button being visible even when element is disabled
+- Made clear button emit null instead of empty string
+- Made input/change event handling a bit more consistent
+- Added defaultValue prop that is for now passed as placeholder to the input field
+
+https://github.com/owncloud/owncloud-design-system/pull/1636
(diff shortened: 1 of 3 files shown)


pascalwengerter (commented):
LGTM and thanks for taking care, would like to get a final approval from @kulmann before merging

kulmann (approved):
LGTM 👍

PR #1639, opened 2021-09-07
Make accessible description on OcAvatars optional

kulmann:
Make accessible description on OcAvatars optional

Also fixed imports of the OcAvatarItem in more specific items on the way.


paulcod3 (approved):
LGTM 👍

pascalwengerter (changes requested):
The example needs each of the OcAvatars components wrapped in a single div / separated by a `<h3>` (or sth) so they show up on separate "lines" and not glued to each other on one line

pascalwengerter:
> The example needs each of the OcAvatars components wrapped in a single div / separated by a `<h3>` (or sth) so they show up on separate "lines" and not glued to each other on one line

Otherwise good to go

pascalwengerter (approved):

PR #1644, opened 2021-09-09
Fix avatar icon alignment

pascalwengerter:
## Description
Moved from a fragile, absolute placement via margin to flexbox-based centering


kulmann (approved):
Awesome! Thanks for the attention to detail 😍

PR #1661, opened 2021-09-19
feat: add v-calendar

LukasHirt:
Bring v-calendar as datepicker and remove vue-datetime


src/components/datepicker/OcDatepicker.vue
```diff
@@ -0,0 +1,52 @@
...
+  release: "1.0.0",
+
+  components: { DatePicker },
+
+  inheritAttrs: true,
+}
+</script>
+
+<style lang="scss">
+.oc-datepicker .vc-highlight {
```
pascalwengerter, line 28: Probably needs some sort of background color styling via CSS variables, too :)

LukasHirt: Not sure if I get you. Do you mean creating a new variable for this background?

package.json
```diff
@@ -113,8 +113,8 @@
     "tippy.js": "^6.3.1",
     "uikit": "3.5.16",
     "url-loader": "^4.1.1",
+    "v-calendar": "^2.3.2",
     "vue": "^2.6.11",
-    "vue-datetime": "^1.0.0-beta.10",
```
pascalwengerter, line 117: I think there's a bunch of `vue-datetime` dependencies we can get rid of, too (e.g. weekstart). Could you take a look and update them accordingly?

changelog/unreleased/change-bring-v-calendar
```diff
@@ -0,0 +1,6 @@
+Change: add v-calendar as our datepicker component
+
+We've added `v-calendar` dependency so that we can use it as our datepicker component.
```
pascalwengerter, line 3:
```suggestion
We've replaced the currently used `vue-datetime` library with `v-calendar`, which will get used in our (now updated) datepicker component.
```

changelog/unreleased/change-bring-v-calendar
```diff
@@ -0,0 +1,6 @@
...
+By doing so, we are deprecating the old datepicker library that we use, `vue-datetime`.
```
pascalwengerter, line 4:
```suggestion
```

src/components/datepicker/OcDatepicker.vue
pascalwengerter: Let me check how theming affects this component before I get back to you on this!

pascalwengerter (commented):
Like the direction this is going ⏳

src/components/datepicker/OcDatepicker.vue
```diff
@@ -0,0 +1,52 @@
...
+      <slot :inputValue="inputValue" :togglePopover="togglePopover" :hidePopover="hidePopover" />
+    </template>
+  </date-picker>
+</template>
+
+<script>
+import DatePicker from "v-calendar/lib/components/date-picker.umd"
+
+/**
+ * Datepicker component based on [v-calendar](https://vcalendar.io/)
```
kulmann, line 14: Could you make it a little bit more obvious that the documentation for the props is external? Like you already did with OcSelect? There it is this text: `For detailed documentation (props, slots, events, etc.), please visit https://vue-select.org/`

src/components/datepicker/OcDatepicker.vue
kulmann: The state like it is now already helps for unblocking development in web, right? Let's merge this, make an rc-release to unblock work in `web`, and do the theming support in a separate ODS PR, ok?

kulmann (commented):
Love it!
Please check out Pascals proposals for the changelog item. Code itself is good.

I'm afraid that the component will not help with accessibility though. ;-) Might be a good idea to enforce `aria-hidden="true"`. Have a look at the screenshot to see what a screenreader reads out when the datepicker is focussed.

Screenshot 2021-09-21 at 21 26 00

src/components/datepicker/OcDatepicker.vue
pascalwengerter: Sure, could be (would also help removing UiKit 😬 )

package.json
pascalwengerter: Removed weekstart, luxon seemed to come with vue-datetime back in the days as well but we're using it outside of the datepicker by now

src/components/datepicker/OcDatepicker.vue
pascalwengerter: @kulmann added it, please review and feel free to merge this PR

kulmann (approved):

PR #1744, opened 2021-10-29
A11y fix sufficient text contrast

stevefrenzel:
## Description
Higher contrast ratio for certain links and buttons.

## Related Issue
- Works towards https://github.com/owncloud/web/issues/5384

## How Has This Been Tested?
- Determined the correct contrast ratio with this contrast checker: https://webaim.org/resources/contrastchecker/
- Tested the components in ODS and web project

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/OcBreadcrumb.vue
+++ b/src/components/OcBreadcrumb.vue
@@ -153,7 +153,6 @@ export default {
     > li span:hover,
     > li button:hover {
       color: var(--oc-color-text-default);
-      text-decoration: underline;
     }
   }
 


src/components/__snapshots__/OcModal.spec.js.snap
```diff
@@ -1,9 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`OcModal displays input 1`] = `
-<div aria-labelledby="oc-modal-title" class="oc-modal-background">
```
pascalwengerter, line 4: Not sure how this change came here - is the snapshot broken on the `a11y-updates` branch?

src/components/OcBreadcrumb.vue
```diff
@@ -143,12 +143,17 @@ export default {
...
+        text-decoration: none;
+      }
     }
 
     > li a:hover,
     > li span:hover,
     > li button:hover {
-      color: var(--oc-color-swatch-primary-default);
+      color: var(--oc-color-text-default);
+      text-decoration: underline;
```
pascalwengerter, line 156: This would also set a hover for a `span` element. Not sure if we want this, what was the idea behind it?

src/styles/theme/oc-text.scss
```diff
@@ -13,7 +13,7 @@ a {
   color: var(--oc-color-swatch-primary-default);
 
   &:hover {
-    color: var(--oc-color-swatch-primary-hover);
+    color: var(--oc-color-swatch-brand-hover);
```
pascalwengerter, line 16: You modify the `swatch-primary-hover` (from my understanding) in the color.yaml file but then change the used variable. Does that mean we don't need the adjustment for the token?

pascalwengerter (changes requested):

src/components/__snapshots__/OcModal.spec.js.snap
pascalwengerter: Nevermind, fixed the outdated snapshot and pushed to `a11y-updates` (= target branch)

src/styles/theme/oc-text.scss
kulmann: good spot @pascalwengerter 
@stevefrenzel please stay within the same color group if possible. text color from `primary` group and hover color from `brand` group would result in a theming mess, semantically.

src/components/OcBreadcrumb.vue
stevefrenzel: This was already implemented, I just changed the color reference to meet a11y requirements. I also added an underline on hover which was not requested, but makes it a bit more obvious it's a clickable link. No problem to remove it again! 🙂

The last `span` element inside a `li` element will have no text decoration or any other hover effect, see line 147 in the same file.

pascalwengerter:
Pretty sure the linked issue needs further adjustments around the "New" button, but merged this one for now

PR #1753, opened 2021-11-02
Limit emit select to single file

pascalwengerter:
## Description
Needed for batch actions in contextmenu in web

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Tests

## Checklist:
- [X] Code changes
- [X] Unit tests added

## Open tasks:
- [ ] Failing unit test (behavior works fine when testing it manually)

### DIFF
--- a/src/components/molecules/OcTable/OcTable.spec.js
+++ b/src/components/molecules/OcTable/OcTable.spec.js
@@ -206,6 +206,18 @@ describe("OcTable", () => {
     expect(wrapper.html().indexOf("data-file-id")).toBeGreaterThan(-1)
   })
 
+  it("emits contextmenu-clicked event upon right click on table row", async () => {
+    const wrapper = shallowMount(Table, {
+      propsData: {
+        fields,
+        data,
+        highlighted: [],
+      },
+    })
+    await wrapper.find(".oc-tbody-tr").trigger("contextmenu")
+    expect(wrapper.emitted().contextmenuClicked.length).toBe(1)
+  })
+
   it("enable dragDrop should enable draggable on rows", () => {
     const wrapper = shallowMount(Table, {
       propsData: {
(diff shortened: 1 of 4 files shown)


kulmann (approved):
👏

kulmann:
@pascalwengerter I think that the selection model might not be tracked correctly from within the tests. Quickly debugged the two `emits no select event if target row is within selected items` tests:
- the first one passes because the `showContextMenu` handler is not called at all (which has the same effect as not emitting a `select` event)
- the second one fails because the `selection` model is always an empty array.

src/components/molecules/OcTable/OcTable.spec.js
```diff
@@ -206,6 +206,18 @@ describe("OcTable", () => {
...
 
+  it("emits contextmenu-clicked event on table row", () => {
+    const wrapper = shallowMount(Table, {
+      propsData: {
+        fields,
+        data,
+        highlighted: [],
+      },
+    })
+    wrapper.vm.$emit("contextmenuClicked")
```
paulcod3, line 217: I'm not sure if this test makes much sense since you emit and event through test and check if it was committed.

kulmann: True. I think in the past it was on a table row and checked if it bubbled up to the table. That didn't work out and the current state vendor-tests the test utils. 😁

pascalwengerter: Meeeh, let me remove it, then

pascalwengerter: Well found out `wrapper.find(".selector").trigger('contextmenu')` exists, switching to that and triggering on a table row makes the test pass and _actually tests business logic_ ™️

paulcod3 (approved):
LGTM thx for taking care :) 👍

pascalwengerter:
> @pascalwengerter I think that the selection model might not be tracked correctly from within the tests. Quickly debugged the two `emits no select event if target row is within selected items` tests:
> 
>     * the first one passes because the `showContextMenu` handler is not called at all (which has the same effect as not emitting a `select` event)
> 
>     * the second one fails because the `selection` model is always an empty array.

Fixed by 1) setting the selection upon mount explicitly and 2) triggering a "real mocked" context-menu click on the row

PR #1765, opened 2021-11-08
Add grouping functionality to OcTableFiles

elizavetaRa:
## Description
Grouping as well as preview functionality for the tables. 

Setup is done with the property "groupingSettings" . Following settings are possible:
- groupingFunctions: Object with keys as grouping options names and functions that get a table data row and return a group name for that row. The names of the functions are used as grouping options. 

- groupingBy: must be either one of the keys in groupingFunctions or 'None'. If not set, default grouping will be 'None'.
- ShowGroupingOptions: boolean value for showing or hinding the select element with grouping options above the table.
- PreviewAmount: Integer value that is used to show only the first n data rows of the table.object

Screenshot from 2021-09-08 10-47-39

## Motivation and Context
UX-friendly option to make the table more

## Types of changes
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- Tests

### DIFF
--- a/src/components/molecules/OcTable/OcTable.vue
+++ b/src/components/molecules/OcTable/OcTable.vue
@@ -1,82 +1,234 @@
 <template>
-  <table v-bind="extractTableProps()">
-    <oc-thead v-if="hasHeader">
-      <oc-tr class="oc-table-header-row">
-        <oc-th
-          v-for="(field, index) in fields"
-          :key="`oc-thead-${field.name}`"
-          v-bind="extractThProps(field, index)"
-          @click.native="$emit(constants.EVENT_THEAD_CLICKED, field)"
-        >
-          <span
-            v-if="field.headerType === 'slot'"
-            :key="field.name + 'Header'"
-            class="oc-table-thead-content"
+  <div>
+    <div
+      v-if="groupingSettings && groupingAllowed && groupingSettings.showGroupingOptions"
+      class="oc-pb-m"
+    >
+      <div class="oc-docs-width-small" style="display: inline">
+        <label class="oc-mx-s">Group By:</label>
+      </div>
+      <div class="oc-docs-width-medium" style="display: inline-block; width: 250px">
+        <oc-select
+          v-model="selected"
+          :options="['None', ...Object.keys(groupingSettings.groupingFunctions)]"
+          :clearable="false"
+          :searchable="false"
+        />
+      </div>
+    </div>
+    <table v-bind="extractTableProps()">
+      <oc-thead v-if="hasHeader">
+        <oc-tr class="oc-table-header-row">
+          <oc-th
+            v-for="(field, index) in fields"
+            :key="`oc-thead-${field.name}`"
+            v-bind="extractThProps(field, index)"
+            @click.native="clickedField(field)"
           >
-            <slot :name="field.name + 'Header'" />
-          </span>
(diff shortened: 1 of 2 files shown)


kulmann:
Could you please run `yarn lint:eslint --fix` please to make eslint happy?

kulmann:
As discussed yesterday we agreed to add tests after merge in a separate PR.

src/components/molecules/OcTable/OcTable.vue
```diff
@@ -1,82 +1,228 @@
...
+            <template v-else>
+              {{ item[field.name] }}
+            </template>
+          </oc-td>
+        </oc-tr>
+      </oc-tbody>
+
+      <!-- Collapsibles -->
+
+      <tbody
```
kulmann, line 90: Please use `oc-tbody`

src/components/molecules/OcTable/OcTable.vue
```diff
@@ -1,82 +1,228 @@
...
+        v-for="(item, index) in groupedItems"
+        v-else-if="groupingAllowed && selected !== 'None'"
```
kulmann, line 92: It's considered bad practice to have `v-for` and `v-(else-)if` on the same element. Only acceptable situation is when you want to skip certain elements of the `v-for`. In this case You'd execute the `v-for` with all the grouped items, just to skip it because of the `v-else-if` if grouping is not allowed.

src/components/molecules/OcTable/OcTable.vue
```diff
@@ -1,82 +1,228 @@
...
-      <oc-tr class="oc-table-header-row">
-        <oc-th
-          v-for="(field, index) in fields"
-          :key="`oc-thead-${field.name}`"
-          v-bind="extractThProps(field, index)"
-          @click.native="$emit(constants.EVENT_THEAD_CLICKED, field)"
+  <div>
+    <div
+      v-if="groupingSettings && groupingAllowed && groupingSettings.showGroupingOptions"
+      style="padding-bottom: 20px"
```
kulmann, line 5: Please use padding helper classes as documented in https://owncloud.design/#/Utilities/Padding

kulmann (commented):
More detailed review will follow, but please fix the linter first. ;-)

src/components/molecules/OcTable/OcTable.vue
elizavetaRa: I tried to put `<template v-else-if...>` wrapper around but for some reason the code after the template gets inactive

pascalwengerter:
Since we've extracted the OcTableFiles to https://github.com/owncloud/web/ in #1817 parts of this PR (the grouping, namely) needs to be ported as well. Please open a separate PR for styling fixes & new icons

diocas:
@pascalwengerter Lisa's going to update this PR, so I think we could keep this open.
It's true that OcTableFiles is no longer here, but we still need changes ini OcTable. If I don't have time to do it myself today, I hope she will be able to do it tomorrow.

pascalwengerter:
Dang my mistake. I'll reopen, but since we're moving sorting functionality out of the ODS too (#1825) let's see if we need this PR, after all :)

PR #1780, opened 2021-11-11
Update NodeJS in CI to v16

pascalwengerter:
## Description
Node v16 is now the official LTS, let's see what CI says!


fschade (approved):
🤖

PR #1787, opened 2021-11-18
Add tooltips to relative dates in files table

JammingBen:
## Description
The following dates are affected: modify date, share date and deleted date.

This goes together with the Web PR https://github.com/owncloud/web/pull/6037

## Related Issue

- Related to https://github.com/owncloud/web/issues/5672

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/enhancement-add-accentuated-property
+++ b/changelog/unreleased/enhancement-add-accentuated-property
@@ -6,7 +6,7 @@ Relative dates like "1 day ago" now have a tooltip that shows the absolute date.
 * share date
 * delete date
 
-The tooltip is a11y-compliant and can be reached via tab.
+The tooltip is a11y-compliant and can be reached via keyboard navigation.
 
 https://github.com/owncloud/owncloud-design-system/pull/1787
 https://github.com/owncloud/web/issues/5672
--- a/changelog/unreleased/enhancement-add-accentuated-property
+++ b/changelog/unreleased/enhancement-add-accentuated-property
@@ -1,0 +1,12 @@
+Enhancement: Relative date tooltips
+
+Relative dates like "1 day ago" now have a tooltip that shows the absolute date. The following dates in the files table are affected:
+
+* modify date
+* share date
+* delete date
+
+The tooltip is a11y-compliant and can be reached via tab.
+
+https://github.com/owncloud/owncloud-design-system/pull/1787
+https://github.com/owncloud/web/issues/5672
--- a/changelog/unreleased/enhancement-add-accentuated-property
+++ b/changelog/unreleased/enhancement-add-accentuated-property
@@ -6,7 +6,5 @@ Relative dates like "1 day ago" now have a tooltip that shows the absolute date.
 * share date
 * delete date
 
-The tooltip is a11y-compliant and can be reached via keyboard navigation.
-
 https://github.com/owncloud/owncloud-design-system/pull/1787
 https://github.com/owncloud/web/issues/5672
--- a/changelog/unreleased/enhancement-add-accentuated-property
+++ b/changelog/unreleased/enhancement-add-accentuated-property
@@ -1,4 +1,4 @@
-Enhancement: Relative date tooltips
+Enhancement: Relative date tooltips in the OcTableFiles component
 
 Relative dates like "1 day ago" now have a tooltip that shows the absolute date. The following dates in the files table are affected:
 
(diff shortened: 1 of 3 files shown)


changelog/unreleased/enhancement-add-accentuated-property
```diff
@@ -0,0 +1,12 @@
+Enhancement: Relative date tooltips
```
pascalwengerter, line 1: Minor nitpick: Wouldn't hurt to mention the OcTableFiles component here to make clear where the relative date tooltips are added

pascalwengerter (approved):
Works like a charm for me 🕙

changelog/unreleased/enhancement-add-accentuated-property
```diff
@@ -0,0 +1,12 @@
...
+
+Relative dates like "1 day ago" now have a tooltip that shows the absolute date. The following dates in the files table are affected:
+
+* modify date
+* share date
+* delete date
+
+The tooltip is a11y-compliant and can be reached via tab.
```
pascalwengerter, line 9:
```suggestion
The tooltip is a11y-compliant and can be reached via keyboard navigation.
```

fschade (approved):
LGTM 🚀

src/components/templates/OcTableFiles/OcTableFiles.vue
```diff
@@ -67,6 +67,27 @@
     <template #size="{ item }">
       <oc-resource-size :size="item.size" />
     </template>
+    <template #mdate="{ item }">
+      <span
+        v-oc-tooltip="formatDate(item.mdate)"
```
kulmann, line 72: @marcus-herrmann could you look into this? I see a conflict of interest here. We're trying to add the absolute date as a tooltip to the relative date (something like `3 hours ago` as a rendered date in the table cell and something like `24.11.2021 09:09` in the tooltip).
The `v-oc-tooltip` directive creates a tooltip with tippy.js on focus and on hover. That's why Jannik gave it a tabindex. The tooltip is by design `aria-hidden="true"` because we needed to decide individually if we need an aria-label or aria-describedby, depending on the context (override vs. enhance accessible content).

Two questions about this:
1. the span is not an interactive element. Is it allowed at all, regarding a11y, to make it focusable then? For keyboard navigation it would be cool, so that you can see the tooltip on focus. But for screen readers it might suggest that it can be interacted with.
2. Do you have an idea how to provide the absolute date in an optional way? I don't think that it's a good idea to give the relative and absolute date at the same time "in you face" to a screen reader user via accessible content + aria-descibredby. Would you recommend an aria-label with combined information in one string, like `3 hours ago (24.11.2021 09:09)`? Or do you have a better idea?

kulmann (commented):
Code looks good to me, but please wait for a response from @marcus-herrmann regarding a11y compliance.

src/components/templates/OcTableFiles/OcTableFiles.vue
marcus-herrmann: Building WCAG/BITV-Test compliant tooltips is really hard (I have to look into tippy.js if this provides an compliant experience).

- don't make something `aria-hidden=true` focussable (if I got that right). Like you are assuming, making something focussable that has no interactive role is irritating at least. On top of that, just triggering it on focus (and de-triggering on blur) would be a violation against WCAG 1.4.13. I would recommend to use either the compliant but less ideal `title` attribute or a more robust pattern, like a toggletip: https://inclusive-components.design/tooltips-toggletips/
- Hm, your suggestion here looks like the sweet spot between the two options

JammingBen: Okay, so, as far as I understand we could:

1) keep the tooltip itself `aria-hidden="true"`
2) add `aria-label` to the span element with something like `3 hours ago (24.11.2021 09:09)`
3) keep the span element focusable (I think the question remains: is that allowed on a "non-interactive" span?)

kulmann: 1. tooltip itself already is `aria-hidden="true"` - that's build into the directive
2. yes, but please keep in mind that the original accessible content **must** be in the aria-label content. You must never lose information from the original accessible content.
3. well, that's my issue. If the fact that it's focusable has no value for a screen reader user then it will be irritating for a screen reader user that it's focusable.

marcus-herrmann: I'm not sure if I'm getting this right, but `aria-label`s does not work on elements that have no role like span. Plus, why are tooltips hidden to assistive technology in the first place?

PR #1795, opened 2021-11-23
Minor code style and docs improvements

kulmann:
Found some small chances for improvement in ODS...

### DIFF
--- a/changelog/unreleased/change-remove-unused-props
+++ b/changelog/unreleased/change-remove-unused-props
@@ -1,0 +1,5 @@
+Change: Remove unused props for unstyled components
+
+We removed the `stopClassProgragation` property in two components, which resulted in unstyled components before but was unused.
+
+https://github.com/owncloud/owncloud-design-system/pull/1795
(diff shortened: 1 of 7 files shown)


pascalwengerter (commented):
Perhaps could use a changelog item since you removed the `stopClassPropagation` in two components, but wouldn't bother with a breaking change (unless we decide to remove UiKit and maybe streamline the form inputs before the next release). Other than that LGTM, thanks for taking care!

kulmann:
> Perhaps could use a changelog item since you removed the `stopClassPropagation` in two components, but wouldn't bother with a breaking change (unless we decide to remove UiKit and maybe streamline the form inputs before the next release). Other than that LGTM, thanks for taking care!

Thanks, I added a changelog item. Indeed a `change` item - although it seems to be not that important. If we don't manage to remove UiKit until the next release we could think about reverting stuff in the release branch...

pascalwengerter (approved):
In favor of merging this one and then going for UiKit soon-ish too so we can have a bunch of `change` items in one release

PR #1798, opened 2021-11-25
Add padding size for OcDrop

kulmann:
## Description
We've added a `paddingSize` property to the OcDrop component for specifying a dedicated padding of the div that wraps the content slot. It defaults to "medium" and also allows to remove the padding entirely.

## Motivation and Context
Being able to remove the padding of an OcDrop where it's not needed / ugly (e.g. role selection for shares in web).

## How Has This Been Tested?

- added unit tests for padding size to class transformation

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/templates/OcTableFiles/OcTableFiles.vue
+++ b/src/components/templates/OcTableFiles/OcTableFiles.vue
@@ -98,6 +98,7 @@
           :popper-options="popperOptions"
           mode="click"
           close-on-click
+          padding-size="remove"
           @click.native.stop.prevent
         >
           <!-- @slot Add context actions that open in a dropdown when clicking on the "three dots" button -->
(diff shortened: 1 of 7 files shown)


paulnebr (approved):
LGTM 👍

src/components/templates/OcTableFiles/OcTableFiles.vue
```diff
@@ -98,6 +98,7 @@
           :popper-options="popperOptions"
           mode="click"
           close-on-click
+          padding-size="remove"
```
pascalwengerter, line 101: @kulmann do you remember why this has been done? Now the files list contextmenu has no padding and we may have to release ODS 11.3.1 for the web release 😬

kulmann: Yes, it's irritating for users that the context menu has a padding on its own, which makes some area around the menu items non-clickable. Ideally we should have the full menu item component area clickable in the future, this works towards it a little bit already.

PR #1831, opened 2021-12-15
[tests-only] refactored try catch use cases inside unit tests

kiranparajuli589:
## Description
To assert the errors thrown, basic `try & catch` implementation was done before.
But this can be replaced using `toThrow` fn from `expect` where we can assert the expected error message thrown.
makes the test code look far better.

## How Has This Been Tested?
- locally
- :robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


pascalwengerter (approved):
LGTM, CI is happy too. We can eventually delete those tests though since we're testing vendor code here

PR #1825, opened 2021-12-12
Extract sort handling from OcTable

dschmidt:
## Description
The sorting depends on the structure of resources which do not exist in ODS but only in web-app-files.
Moreover, we need to sort before pagination and as pagination is not handled in `OcTable` either, we need to handle sorting externally as well.
This has the additional benefit that this prepares us better for retrieving server sorted and paginated resources.

This is a draft and as such has a few rough edges.
For instance sorting has been added to the first example of `OcTable`, should we get this close to merging this would deserve its own section obviously :) 

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation added/updated

## Open tasks:

- [ ] ...

PR #1839, opened 2021-12-20
Allow sort direction defaults from field

kulmann:
## Description
Allow a sort direction default coming from the field definition instead of just assuming it. This is important because the default sort direction for the `name` column (`ASC`)  is different from the default sort direction for the `mdate` column (`DESC`). We have no way of assuming this. Should be provided instead.

This is considered a continuation of https://github.com/owncloud/owncloud-design-system/pull/1825 so no additional changelog item provided.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


pascalwengerter (approved):
Bonus points for the cleaned-up unit tests, LGTM!

PR #1826, opened 2021-12-13
Switch iconset to remix icons

lookacat:
## Description
Switch from our rather random collection of icons to Remix icons

## Related Issue
- Fixes #1348

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added/updated

## Open tasks:

- [ ] ...

PR #1846, opened 2021-12-27
Add svg icon for spaces

JammingBen:
image


pascalwengerter (approved):
LGTM, we need to check how we handle the transfer in #1826 though :)

PR #1919, opened 2022-02-02
Switch OcDrop background to muted, add rounded corners helper class

pascalwengerter:
## Description
Part of the redesign works

### DIFF
--- a/src/styles/theme/helper.scss
+++ b/src/styles/theme/helper.scss
@@ -81,3 +81,27 @@ hr {
   overflow: visible;
   text-align: inherit;
 }
+
+.oc-rounded {
+  border-radius: 5px;
+}
+
+.oc-rounded-top {
+  border-top-left-radius: 5px;
+  border-top-right-radius: 5px;
+}
+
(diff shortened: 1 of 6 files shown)


src/styles/theme/helper.scss
```diff
@@ -81,3 +81,17 @@ hr {
   overflow: visible;
   text-align: inherit;
 }
+
+.oc-rounded {
+  border-radius: 5px;
+}
+
```
kulmann, line 88: What do you think about two additional classes, `oc-rounded-top` and `oc-rounded-bottom` for only setting the 5px border radius top or bottom? I guess that there are situations where you know that you only need the rounded borders top or bottom.

kulmann: And in such cases it would be nicer to have a dedicated class for it instead of applying one class for rounded corners and then another class for removing rounded corners on one end again.

pascalwengerter: Yeah true, let me fix that in a minute

kulmann (approved):

PR #1936, opened 2022-02-07
Clean up component release info

pascalwengerter:
Some components on https://owncloud.design/ are being listed as either `In review` or with missing release info. This PR fixes it (couldn't find all the relevant release info for e.g. OcList though)


kulmann (approved):
whoopsie. thank you!

PR #1941, opened 2022-02-08
Redesign: add hover subline to breadcrumbs

lookacat:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/molecules/OcBreadcrumb/OcBreadcrumb.vue
+++ b/src/components/molecules/OcBreadcrumb/OcBreadcrumb.vue
@@ -207,6 +207,7 @@ export default {
     > li span:not([aria-current="page"]):hover,
     > li button:hover {
       color: var(--oc-color-swatch-primary-default);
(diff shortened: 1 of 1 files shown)


pascalwengerter (commented):
Not sure if we want to drop the hover color on breadcrumbs in favor of underline text, or do we want both @tbsbdr @kulmann ?

tbsbdr:
I would not drop the hover color completely, as it looks a little bit nicer to have it. But as we should avoid to rely only on color to communicate a state, underlining is imo more important (will not always be possible i guess). hover-color is nice to have.
image

pascalwengerter (approved):

PR #1958, opened 2022-02-14
Redesign FilesTable

lookacat:
## Description
See https://github.com/owncloud/web/issues/6207

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


src/components/organisms/OcResource/OcResource.vue
```diff
@@ -205,6 +215,9 @@ export default {
   },
 
   methods: {
+    hover() {
```
kulmann, line 218: remove dev leftover

src/components/atoms/OcCheckbox/OcCheckbox.vue
```diff
@@ -133,18 +133,17 @@ export default {
...
-  background-position: 50% 50%;
-  background-repeat: no-repeat;
-  background-size: 80%;
-  border: 1px solid var(--oc-color-input-border);
-  border-radius: 3px;
+
+  background-position: 50% 50% !important;
+  background-repeat: no-repeat !important;
+  border: 2px solid var(--oc-color-input-border);
+  border-radius: 5px;
```
kulmann, line 142: you can extend the `.oc-rounded` class instead of setting a border-radius manually.

changelog/unreleased/enhancement-redesign-filestable
```diff
@@ -0,0 +1,7 @@
+Enhancement: Redesign FilesTable
```
kulmann, line 1: I'd say `Enhancement: Redesign files table related components` because the ODS doesn't know anything about a files table.

kulmann (changes requested):
Found some small things.
Regarding the new color variable I'm a bit confused now. I thought that the `background-secondary` was supposed to be another surface color variable with the color value that was previously given to the highlight color. And that you wanted to come up with a completely new color for the already existing `background-highlight` color. 🤔 could you resolve my confusion?

changelog/unreleased/enhancement-redesign-filestable
lookacat: yes sounds better

src/components/organisms/OcResource/OcResource.vue
lookacat: damn didnt see it :D

src/components/atoms/OcCheckbox/OcCheckbox.vue
```diff
@@ -133,18 +133,16 @@ export default {
...
-  border: 1px solid var(--oc-color-input-border);
-  border-radius: 3px;
+
+  background-position: 50% 50% !important;
+  background-repeat: no-repeat !important;
+  border: 2px solid var(--oc-color-input-border);
   display: inline-block;
   overflow: hidden;
   vertical-align: middle;
+  background: transparent;
```
kulmann, line 145:
```suggestion
  background-color: transparent;
```

src/tokens/ods/color.yaml
```diff
@@ -77,6 +77,8 @@ color:
   background:
     default:
       value: rgb(255, 255, 255)
+    secondary:
+      value: rgb(236, 236, 236)
```
kulmann, line 81: I'm still confused. You wanted to set the current `background-highlight` color value here for the `background-secondary` and then come up with a new color value for `background-highlight` which better visualizes a highlighted state.
Or in other words: We want to keep the color value for the OcDrop background, but the variable name that was used there is semantically questionable. `background-secondary` would be a better naming for the OcDrop background, but the actual color of the OcDrop is supposed to stay as it is.

kulmann (changes requested):
- Still not happy with the usage/changes regarding color variables (see detailed comment)
- Sorting icons in table header are only visible on hover in dark mode. In light mode not even on hover. In both modes not permanently for the currently sorted column.

But:
- OcCheckbox looks really good again 🥳

src/tokens/ods/color.yaml
kulmann: discussed in a call that the code it's good like it is code wise 👍

kulmann (approved):
Looks good to me 👍  Discussed in a call that you'll add the `background-hover` color token for highlighting in the files table, that's also good to get in.

PR #1969, opened 2022-02-22
extend modal focus-trap

fschade:
* add modal option to disable focus trap
* add modal option to define which child element should be focused initially


src/components/organisms/OcModal/OcModal.vue
```diff
@@ -277,10 +295,9 @@ export default {
       this.userInputValue = value
     },
     getInitialFocusRef() {
-      if (this.hasInput) {
-        return this.$refs.ocModalInput
-      }
-      return this.$refs.ocModal
+      console.log(111, this.focusTrapInitial)
```
lookacat, line 298: dev leftover?

lookacat (approved):
one dev leftover but otherwise good :)

src/components/organisms/OcModal/OcModal.vue
fschade: 🦀  yup, thanks

PR #1970, opened 2022-02-22
Revert "extend modal focus-trap"

fschade:
Reverts owncloud/owncloud-design-system#1969


lookacat (approved):
Compared with old file and all changes reverted, LGTM 👍🏼

PR #1978, opened 2022-02-24
Only allow hover in OcTable when enabled

pascalwengerter:
Seems to have gone unnoticed in https://github.com/owncloud/owncloud-design-system/pull/1958, but suddenly we had a hover on the OcTable footer. This kinda reverts the changes to only allow hover on none-footer-cells and when hovering is enabled for the component via its props


lookacat (approved):
LGTM, good finding!

lookacat (approved):
LGTM, good finding!

lookacat (approved):
LGTM, good finding!

PR #2019, opened 2022-03-14
Sprint 16 design fixes

lookacat:
See https://github.com/owncloud/web/issues/6492
and https://github.com/owncloud/web/pull/6584

### DIFF
--- a/src/components/atoms/OcButton/OcButton.vue
+++ b/src/components/atoms/OcButton/OcButton.vue
@@ -132,6 +132,7 @@ export default {
     $_ocButton_buttonClass() {
       const classes = [
         "oc-button",
+        "oc-rounded",
         `oc-button-${getSizeClass(this.size)}`,
         `oc-button-justify-content-${this.justifyContent}`,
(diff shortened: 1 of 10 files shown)


src/components/atoms/OcButton/OcButton.vue
```diff
@@ -258,11 +258,11 @@ export default {
   align-items: center;
   border: 1px solid;
   border-color: transparent;
-  border-radius: 3px;
+  border-radius: 5px;
```
kulmann, line 261: Since you're changing it to our default border radius size, could you extend `oc-rounded` instead? Same for the other two lines where you changed the border radius to `5px`. I'd like to avoid writing the same hardcoded value throughout the code base and instead reuse / extend existing classes.

lookacat: yes good finding

lookacat: tried to use it wherever it was possible, for :hover i couldn't

kulmann (approved):

Issue #1952, opened 2022-02-10
Give primary button the new gradient style


kulmann:
Our `primary` button variation doesn't have the gradient style, yet. It's only applied via a css class to some handpicked buttons in web. Let's restyle the `primary` variation of the button according to the re-design requirements.

## wrong / current
<img width="1113" alt="Screenshot 2022-02-10 at 10 08 52" src="https://user-images.githubusercontent.com/3532843/153374882-6d3f5f99-e60d-4226-bbbd-c6928a61db7f.png">

## new design
<img width="191" alt="Screenshot 2022-02-10 at 10 12 04" src="https://user-images.githubusercontent.com/3532843/153374987-3e0df23e-3603-4e79-a1f6-d06d81565a67.png">

PR #2036, opened 2022-03-21
Small fixes

### LINKED ISSUE #2030: [OcSelect] Background is not completely invisible
At least in dark mode the background of the OcSelect component does not seem to be completely invisible.
When not expanded, you see it on the top corners but also the bottom corners.
If you expand it, it's fixed for the bottom corners.

### Steps to reproduce
Look at settings (e.g. https://host.docker.internal:9200/settings)

### Screenshots (if appropriate):

Screenshot_20220318_183743

Screenshot_20220318_183725

#### Browser log
```

Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log 
c) ...
```

pascalwengerter:
## Description
- Fixes #2030
- Fixes raw buttons not respecting the fontsize set by button-size prop
- Update dependencies & implicit import in OcTable


kulmann (changes requested):
Results in huge raw buttons everywhere, which feels wrong. see screenshot.

Screenshot 2022-03-21 at 14 31 22

And in context menu it's like this (see "Open folder" vs the other actions):
Screenshot 2022-03-21 at 14 32 15

IMO we need to leave the raw buttons untouched. Side by side buttons of different appearances need to be avoided :-/

kulmann:
> Damn. So #1952 is the way to go here?

yes, I think so

kulmann (approved):

PR #2049, opened 2022-03-28
Redesign _OcGhostElement

lookacat:
## Description
We've redesigned OcGhostElement to use OcResourceIcon and
to display better preview of the items that have been dragged

## Screenshots
- Now:
Darkmode: 3 items selected (folder):
Bildschirmfoto 2022-04-04 um 12 24 39

Darkmode: 3 items selected (file):
Bildschirmfoto 2022-04-04 um 12 27 26

Lightmode 2 items selected:
Bildschirmfoto 2022-04-04 um 12 28 41

- Before
drags3

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/atoms/_OcGhostElement/_OcGhostElement.vue
+++ b/src/components/atoms/_OcGhostElement/_OcGhostElement.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="ghost-element" class="ghost-element">
     <div class="ghost-element-layer1 oc-rounded">
-      <oc-resource-icon :resource="previewItems[0]" />
+      <oc-resource-icon class="oc-p-xs" :resource="previewItems[0]" />
       <div v-if="showSecondLayer" class="ghost-element-layer2 oc-rounded" />
       <div v-if="showThirdLayer" class="ghost-element-layer3 oc-rounded" />
     </div>
--- a/src/components/atoms/_OcGhostElement/_OcGhostElement.vue
+++ b/src/components/atoms/_OcGhostElement/_OcGhostElement.vue
@@ -1,11 +1,12 @@
 <template>
   <div id="ghost-element" class="ghost-element">
-    <div style="position:relative">
+    <div style="position: relative">
       <div v-for="index in iconStackCount" :key="index">
         <oc-resource-icon
           :class="getIconClass(index - 1)"
-          :style="getStyleClass(index -1)"
-          :resource="previewItems[index - 1]">
+          :style="getStyleClass(index - 1)"
+          :resource="previewItems[index - 1]"
+        >
         </oc-resource-icon>
       </div>
     </div>
@@ -29,8 +30,8 @@ export default {
   },
   computed: {
     iconStackCount() {
-      const count = this.previewItems.length;
-      return count > 3 ? 3 : count;
+      const count = this.previewItems.length
+      return count > 3 ? 3 : count
     },
     firstPreviewIcon() {
       const icon = this.previewItems[0].icon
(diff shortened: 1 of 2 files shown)


src/components/atoms/_OcGhostElement/_OcGhostElement.vue
```diff
@@ -20,6 +29,10 @@ export default {
     },
   },
   computed: {
+    iconStackCount() {
+      const count = this.previewItems.length
+      return count > 3 ? 3 : count
```
JanAckermann, line 34: Math.max maybe ;)

lookacat: yes :D right

src/components/atoms/_OcGhostElement/_OcGhostElement.vue
```diff
@@ -1,6 +1,15 @@
 <template>
   <div id="ghost-element" class="ghost-element">
-    <oc-icon :name="firstPreviewIcon" size="large" />
+    <div style="position: relative">
```
lookacat, line 3: no inline styles pls

kulmann:
Could you add before/after screenshots to the PR description? Helpful for anyone that stumbles over this PR and doesn't want to check it out and build it. ;-)

src/components/atoms/_OcGhostElement/_OcGhostElement.vue
JanAckermann: We might also at a variable with a descriptive name for the magic number 3 ;)

src/components/atoms/_OcGhostElement/_OcGhostElement.vue
```diff
@@ -28,10 +41,41 @@ export default {
       return this.previewItems.length
     },
   },
+  methods: {
+    getStyleClass(index) {
```
kulmann, line 45: naming is not ideal, since you return `styles`, not `classes`.

src/components/atoms/_OcGhostElement/_OcGhostElement.vue
lookacat: yes will do so

src/components/atoms/_OcGhostElement/_OcGhostElement.vue
lookacat: yes

src/components/atoms/_OcGhostElement/_OcGhostElement.vue
```diff
@@ -1,6 +1,22 @@
...
+        :resource="iconFolder"
+      />
+      <div v-for="(item, index) in iconStack" :key="index" :style="{ 'z-index': key }">
+        <oc-icon
+          :style="getStyle(index)"
+          class="ghost-element-icon-background"
+          name="resource-type-file"
+          size="large"
+          variation="inverse"
+        />
```
lookacat, line 16: Any idea for improvement welcome: This is a file icon without any cutout rendered behind the actual file icon. This prevents it from looking weird, look for example at the image or text icon. They both have a cutout in the middle which would look bad in this case.

pascalwengerter:
Checked it out with a local link, two things:
- why not put the folder in the front if there is one? It's lower vertically so badly visible if it's in the background
- folder dragitem color has a bad contrast on darkmode (due to opacity, I guess?)

lookacat:
Folder in the foreground looks weirder than in the background, so left it like this

kulmann (changes requested):
Screenshot 2022-04-01 at 14 37 32 (3)

I'm sorry, but this still doesn't look nice 😕 
@tbsbdr what's your opinion here?

lookacat:
same, im still not 100% happy, maybe do it more like dropbox?
Bildschirmfoto 2022-04-04 um 09 41 32

lookacat:
Bildschirmfoto 2022-04-04 um 11 34 08
how about this? @kulmann @tbsbdr

pascalwengerter:
> Bildschirmfoto 2022-04-04 um 11 34 08
> how about this? @kulmann @tbsbdr

Looks pretty dope in my opinion. Could you add screenshots for lightmode (and with a selected folder for both dark&light mode as well?)

kulmann (changes requested):
two more nitpicks:
- folder icon is too big compared to the file type icons. it has close to zero padding inside the  container
- IMO the ghost element can have opacity 1 in your lastest version. looks weird that underlying content is shining through the icons.

lookacat:
> two more nitpicks:
> 
> * folder icon is too big compared to the file type icons. it has close to zero padding inside the  container
> * IMO the ghost element can have opacity 1 in your lastest version. looks weird that underlying content is shining through the icons.

@kulmann Yes folder padding is fixed now, ghost elements cant have custom opacity, its set by the browser (since it uses `event.dragImage`)

kulmann (approved):

PR #2064, opened 2022-04-06
Create OcContextualHelper component

lookacat:
## Description
See https://github.com/owncloud/web/issues/6590
## Related Issue

- Works towards [<issue_link>](https://github.com/owncloud/web/issues/6590)

## Screenshots
Screenshot (10)

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/atoms/OcButton/OcButton.vue
+++ b/src/components/atoms/OcButton/OcButton.vue
@@ -57,6 +57,17 @@ export default {
       default: null,
     },
     /**
+     * When setting the button’s type to a link, use this option to give a give a target.
+     * `_blank, _self, _parent, _top`
+     */
+    target: {
+      type: String,
+      default: null,
+      validator: value => {
+        return value.match(/(_blank|_self|_parent|_top)/)
+      },
+    },
+    /**
      * When setting the button’s type to a router-link, use this option to give a to.
      */
     to: {
@@ -144,6 +155,7 @@ export default {
     additionalAttributes() {
       return {
         ...(this.href && { href: this.href }),
+        ...(this.target && { target: this.target }),
         ...(this.to && { to: this.to }),
         ...(this.type === "button" && { type: this.submit }),
       }
--- a/src/components/atoms/OcContextualHelper/OcContextualHelper.spec.js
+++ b/src/components/atoms/OcContextualHelper/OcContextualHelper.spec.js
@@ -4,8 +4,8 @@ import GetTextPlugin from "vue-gettext"
 
 const localVue = createLocalVue()
 localVue.use(GetTextPlugin, {
-  translations: 'does-not-matter.json',
-  silent: true
+  translations: "does-not-matter.json",
+  silent: true,
 })
 
 describe("OcContextualHelper", () => {
--- a/src/components/atoms/OcContextualHelper/OcContextualHelper.spec.js
+++ b/src/components/atoms/OcContextualHelper/OcContextualHelper.spec.js
@@ -1,0 +1,45 @@
+import OcContextualHelper from "./OcContextualHelper.vue"
+import { createLocalVue, shallowMount } from "@vue/test-utils"
+import GetTextPlugin from "vue-gettext"
+
+const localVue = createLocalVue()
+localVue.use(GetTextPlugin, {
+  translations: 'does-not-matter.json',
+  silent: true
+})
+
+describe("OcContextualHelper", () => {
+  function getWrapperWithProps(props) {
+    return shallowMount(OcContextualHelper, {
+      localVue: localVue,
+      propsData: props,
+      stubs: {
+        OcDrop: true,
+      },
+    })
+  }
+  describe("should use props correctly", () => {
+    it("should set text prop", () => {
+      const wrapper = getWrapperWithProps({ text: "test-my-text" })
+      expect(wrapper.find(".info-text").text()).toBe("test-my-text")
+    })
+    it("should set list prop", async () => {
+      const listValues = ["a-list-value", "b-list-value", "c-list-value"]
+      const wrapper = getWrapperWithProps({ list: listValues })
+      const result = wrapper.find(".info-list").text()
+      listValues.forEach(value => {
+        expect(result).toContain(value)
+      })
+    })
+    it("should set a readMore link", async () => {
+      const wrapper = getWrapperWithProps({ readMoreLink: "owncloud.design" })
+      const attributes = wrapper.find(".info-more-link").attributes()
+      expect(attributes["href"]).toBe("owncloud.design")
+      expect(attributes["target"]).toBe("_blank")
+    })
+    it("should set end-text prop", async () => {
+      const wrapper = getWrapperWithProps({ endText: "test-my-text" })
+      expect(wrapper.find(".info-text-end").text()).toBe("test-my-text")
+    })
+  })
+})
--- a/src/components/atoms/OcContextualHelper/OcContextualHelper.vue
+++ b/src/components/atoms/OcContextualHelper/OcContextualHelper.vue
@@ -1,0 +1,138 @@
+<template>
+  <div class="oc-contextual-helper">
+    <oc-button :id="buttonId" appearance="raw">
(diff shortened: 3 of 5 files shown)


lookacat:
For clarification: Of couse slots would be better in most cases but the idea here is to give all information by props so we can create json files with the content of each helper in web. This would mean almost no markup in web, if we used slots it would create markup that is not relevant for web.

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,128 @@
...
+    },
+  },
+}
+</script>
+
+<style lang="scss">
+.oc-contextual-helper {
+  display: inline-block;
+  .oc-button {
+    vertical-align: middle;
```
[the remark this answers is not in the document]
lookacat: yes, because without it the helper icon will look weird if used in text :)

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <div class="oc-contextual-helper">
+    <oc-button :id="`${buttonId}`" appearance="raw">
```
pascalwengerter, line 3: No need to do string interpolation here since the computed prop already returns a string (?) ;)

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
...
+      <oc-icon name="question" fill-type="line" size="small" />
+    </oc-button>
+    <oc-drop :drop-id="dropId" :toggle="`#${buttonId}`" mode="click" close-on-click>
```
pascalwengerter, line 6: No need to do string interpolation here since the computed prop already returns a string (?) ;)

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
...
+      <div class="info-drop-content">
+        <span class="info-text" v-text="text" />
```
pascalwengerter, line 8: Probably a `<p>` tag is the better choice here

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
...
+        <ul v-if="list.length" class="info-list oc-pl-l">
+          <li v-for="(item, index) in list" :key="index" class="oc-pl-rm">
+            {{ item }}
+          </li>
+        </ul>
+        <span class="info-text-end" v-text="endText" />
```
pascalwengerter, line 14: Same here, should perhaps also be a `<p>` tag

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
...
+        <a v-if="readMoreLink" class="info-more-link" :href="readMoreLink" target="_blank"
+          >Read more</a
```
pascalwengerter, line 16: My previous comment about missing translations has been marked as resolved but I don't see $gettext being used here?

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
...
+        >
+      </div>
+    </oc-drop>
+  </div>
+</template>
+
+<script>
+import uniqueId from "../../../utils/uniqueId"
+import OcButton from "../../atoms/OcButton/OcButton.vue"
+import OcIcon from "../../atoms/OcIcon/OcIcon.vue"
```
pascalwengerter, line 26: You also use the OcDrop component here?

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,129 @@
...
+    <oc-drop :drop-id="dropId" :toggle="`#${buttonId}`" mode="click" close-on-click>
+      <div class="info-drop-content">
+        <span class="info-text" v-text="text" />
+        <ul v-if="list.length" class="info-list oc-pl-l">
+          <li v-for="(item, index) in list" :key="index" class="oc-pl-rm">
+            {{ item }}
+          </li>
+        </ul>
+        <span class="info-text-end" v-text="endText" />
+        <a v-if="readMoreLink" class="info-more-link" :href="readMoreLink" target="_blank"
```
pascalwengerter, line 15: Maybe using another (raw, passive) oc-button here would help with color contrasts through theming etc?

pascalwengerter (changes requested):
Sorry, more findings 🐺

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
lookacat: in this case it is because of the `#`

pascalwengerter: But then why not turn around the computed prop assignment (generate button ID first and then use it with the prefixed `#` for the drop)? Seems cleaner than doing string interpolation within the template

lookacat: I dont get it

src/components/atoms/OcButton/OcButton.vue
```diff
@@ -56,6 +56,13 @@ export default {
       type: String,
       default: null,
     },
+    /**
+     * When setting the button’s type to a link, use this option to give a give a target.
+     */
+    target: {
+      type: String,
+      default: null,
```
pascalwengerter, line 64: Super nitpicky, but shouldn't we only allow the few options available in the HTML standard? https://www.w3schools.com/tags/att_a_target.asp

src/components/atoms/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -0,0 +1,141 @@
...
+  display: inline-block;
+  .oc-button {
+    vertical-align: middle;
+  }
+  .info-drop-content {
+    font-size: var(--oc-font-size-small);
+    color: var(--oc-color-text-default);
+  }
+  .info-list {
+    list-style-position: outside;
```
pascalwengerter, line 85: Isn't this the CSS standard value? Why set it explicitly?

pascalwengerter (commented):
😬

src/components/atoms/OcContextualHelper/OcContextualHelper.spec.js
```diff
@@ -0,0 +1,44 @@
+import Vue from "vue"
+import OcContextualHelper from "./OcContextualHelper.vue"
+import { createLocalVue, shallowMount } from "@vue/test-utils"
+import GetTextPlugin from "vue-gettext"
+
+const localVue = createLocalVue()
+
+Vue.use(GetTextPlugin, { translations: {} })
```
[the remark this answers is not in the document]
lookacat: doesn't make a difference but yes sounds alright ^^

src/components/atoms/OcButton/OcButton.vue
lookacat: theoretically you could use an iframe as an target, but i think we should validate because thats pretty rare.

lookacat:
done :)

src/components/atoms/OcContextualHelper/OcContextualHelper.spec.js
```diff
@@ -0,0 +1,43 @@
+import Vue from "vue"
```
pascalwengerter, line 1: Now you don't need this anymore, right?

lookacat: nope still needed

pascalwengerter (approved):
LGTM 🖖🏽

PR #2137, opened 2022-05-12
Fix tooltips overflowing on long words

javfg:
This PR fixes long words or paths breaking the tooltips in the sidebar:

image

By adding a `word-wrap: break-word` to the tippy content.

image

This along with ellipsizing the paths will remove the scrollbar from the sidebar.

### DIFF
--- a/changelog/unreleased/bugfix-tooltip-break-word
+++ b/changelog/unreleased/bugfix-tooltip-break-word
@@ -1,0 +1,5 @@
+Bugfix: Add word breaking in tooltips
+
+We've added word wrapping to the tippy tooltips so they handle very long paths properly.
+
+https://github.com/owncloud/owncloud-design-system/pull/2137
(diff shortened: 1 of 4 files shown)


pascalwengerter (changes requested):
Could you move the style rules into a separate tooltip.scss or tippy.scss file in `src/styles/theme`? I'm happy to approve&merge, but this does not (only) apply to OcDrop components, but is rather targeted at the tooltip directive :)

pascalwengerter:
Also, please add a changelog item;)

javfg:
@pascalwengerter Thanks for the suggestions! I've made the requested changes.

changelog/unreleased/bugfix-tooltip-break-word
```diff
@@ -0,0 +1,5 @@
+Bugfix: Add word breaking in tooltips
+
+We've added word wrapping to the tippy tooltips so they handle very long paths properly.
```
pascalwengerter, line 3: Technically they handle all long strings properly :P but that should be fine

pascalwengerter (approved):
Thanks!

PR #2149, opened 2022-05-19
OcModal: Add checkbox and secondary button

lookacat:
## Description
We've added an optional checkbox and secondary button to the OcModal

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6996

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/organisms/OcModal/OcModal.vue
+++ b/src/components/organisms/OcModal/OcModal.vue
@@ -34,7 +34,7 @@
             @keydown.enter="confirm"
           />
           <p v-else key="modal-message" class="oc-modal-body-message" v-text="message" />
-          <div v-if="checkbox" class="oc-modal-body-actions oc-flex oc-flex-left">
+          <div v-if="checkboxLabel" class="oc-modal-body-actions oc-flex oc-flex-left">
             <oc-checkbox
               v-model="checkboxValue"
               size="medium"
@@ -51,7 +51,7 @@
               v-text="buttonCancelText"
             />
             <oc-button
-              v-if="buttonSecondary"
+              v-if="buttonSecondaryText"
               class="oc-modal-body-actions-secondary oc-ml-s"
               :variation="buttonSecondaryVariation"
               :appearance="buttonSecondaryAppearance"
@@ -146,14 +146,6 @@ export default {
       default: null,
     },
     /**
-     * Show modal checbox
-     */
-    checkbox: {
-      type: Boolean,
-      required: false,
-      default: false,
-    },
-    /**
      * Modal checkbox label
      */
     checkboxLabel: {
@@ -190,14 +182,6 @@ export default {
       validator: value => {
         return value.match(/(outline|filled|raw)/)
       },
-    },
-    /**
-     * Text of the secondary button
-     */
-    buttonSecondary: {
-      type: Boolean,
-      required: false,
-      default: false,
     },
     /**
      * Text of the secondary button
@@ -552,7 +536,6 @@ export default {
     message="Do you accept our terms of use?"
     button-cancel-text="Decline"
     button-confirm-text="Accept"
-    checkbox="true"
     checkbox-label="I accept the terms of use"
     class="oc-mb-l oc-position-relative"
   />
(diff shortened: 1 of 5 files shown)


src/components/organisms/OcModal/OcModal.vue
```diff
@@ -459,6 +488,16 @@ export default {
...
     </template>
   </oc-modal>
+  <oc-modal
+    icon="information"
+    title="Accept terms of use"
+    message="Do you accept our terms of use?"
+    button-cancel-text="Decline"
+    button-confirm-text="Accept"
+    checkbox="true"
+    checkbox-label="I accept the terms of use"
```
kulmann, line 498: haha, actually very comprehensible, yet annoying use case. Good choice for the docs 👍

src/components/organisms/OcModal/OcModal.vue
```diff
@@ -34,6 +34,9 @@
             @keydown.enter="confirm"
           />
           <p v-else key="modal-message" class="oc-modal-body-message" v-text="message" />
+          <div v-if="checkbox" class="oc-modal-body-actions oc-flex oc-flex-left">
+            <oc-checkbox size="medium" :value="checkboxValue" :label="checkboxLabel" :aria-label="checkboxLabel"/>
```
kulmann, line 38: You're never setting the `checkboxValue`. Needs two way data binding via `v-model="checkboxValue`.
I guess we also want to allow setting an initial value for the checkbox.

kulmann (changes requested):
Why do you need additional props? Is it not feasible to implement a component API that uses `inputType="checkbox"`?

src/components/organisms/OcModal/OcModal.vue
lookacat: yes true

lookacat:
`inputType` seems to be very bound to `oc-text-input`, also would argue that you may want an text input with a checkbox

src/components/organisms/OcModal/OcModal.vue
```diff
@@ -157,6 +191,44 @@ export default {
...
       },
     },
+    /**
+     * Text of the secondary button
+     */
+    buttonSecondary: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
```
dschmidt, line 201: I'm not sure whether I like this switch or not. 

My feeling is that it's a bit odd having to define `buttonSecondary` as `true` if I already provide `buttonSecondaryText`. Of course that makes it a bit easier to toggle, but maybe toggling the secondaryText is easy enough as well.

src/components/organisms/OcModal/OcModal.vue
```diff
@@ -157,6 +191,44 @@ export default {
...
+    /**
+     * Text of the secondary button
+     */
+    buttonSecondaryText: {
```
dschmidt, line 205: What is this good for after all?

I find there's very little semantics in the name `buttonSecondary` could be everything and nothing, currently it's rendered in between confirm and cancel buttons... I feel this might end up being used very inconsistently.

lookacat: I kinda agree with the naming but can't think of anything better, but we need a button besides cancel and confirm as seen in the mockups in this issue https://github.com/owncloud/web/issues/6996

dschmidt: Ah, sorry, missed the link in your description. My bad.

Naming is hard 🤷🏻

lookacat: If anyone comes up with a better name I will change it ^^ would leave it like this for now

src/components/organisms/OcModal/OcModal.vue
```diff
@@ -34,6 +34,14 @@
             @keydown.enter="confirm"
           />
           <p v-else key="modal-message" class="oc-modal-body-message" v-text="message" />
+          <div v-if="checkboxLabel" class="oc-modal-body-actions oc-flex oc-flex-left">
+            <oc-checkbox
+              v-model="checkboxValue"
+              size="medium"
+              :label="checkboxLabel"
+              :aria-label="checkboxLabel"
```
kulmann, line 42: if aria-label and accessible content (in this case: `label`) are the same you don't need the aria-label. You can drop it completely.

kulmann (approved):

PR #2158, opened 2022-05-23
Add root file type icon

javfg:
This PR adds the file icon for [root](https://root.cern.ch/) files.

It would be great if this can be put in the ODS along with the other icons for files from FontAwesome. We've made it so the SVG is very similar to the rest of the file type icons. But if you prefer another approach, please let us know!


kulmann:
Did you create the icon yourself? Can you add a kind of license comment into the svg file?

javfg:
> Did you create the icon yourself? Can you add a kind of license comment into the svg file?

It was created by @elizavetaRa, I just made it black and remove inkscape metadata. I've added a MIT license XML entry. Let me know if that works.

javfg:
Pinging @kulmann. Changes are ready. Thanks for the suggestions!

javfg:
> Please also add an author(s) comment to the svg (github handles). Otherwise the license statement is hard to follow.

@kulmann do you mean adding @elizavetaRa in the `Co-authored-by`?

The rest of changes are applied too.

src/assets/icons/resource-type-root-fill.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512">
+   <!-- MIT License: https://opensource.org/licenses/MIT -->
```
kulmann, line 2: sorry, I meant adding another comment here in the svg file, linking yours and @elizavetaRa github profiles as authors. just something like:
`<!-- author: https://github.com/elizavetaRa -->`

javfg: Ah, that makes sense! I've just added that.

kulmann (approved):

PR #2164, opened 2022-05-24
Add HTML title to the resourceName component

JammingBen:
## Description
We've added the HTML title attribute to the resourceName component in case no tooltip is being displayed.

## Related Issue

- Part of https://github.com/owncloud/web/issues/6776

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-resource-name-html-title
+++ b/changelog/unreleased/enhancement-resource-name-html-title
@@ -1,0 +1,5 @@
+Enhancement: Add HTML title to the resourceName component
+
+We've added the HTML title attribute to the resourceName component in case no tooltip is being displayed.
+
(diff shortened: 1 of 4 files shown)


dschmidt:
Why is this handled via title instead of computing a value for tooltip in this case as well?

JammingBen:
> Why is this handled via title instead of computing a value for tooltip in this case as well?

We don't want our custom tooltip as default because it's very "in your face". Its only purpose is to show additional information if necessary (the path in this case).

The HTML tooltip is way more subtle, hence we decided to use it in case our custom tooltip is not showing. The main use case is when a file name gets truncated because of its length, but the user still wants to know the full name.

dschmidt:
Ok, I guess 😁

dschmidt (approved):

PR #2200, opened 2022-06-13
Implement an indeterminate state for the progress bar

JammingBen:
## Description
We've added an indeterminate state to the progress bar.

I've decided to build the indeterminate state myself (okay, with quite some help from StackOverflow :D) because the integration of Vuetify is a bit more complex than I anticipated. We can still do this later if neccessary.

## Related Issue
Needed for https://github.com/owncloud/web/issues/7105

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/src/components/atoms/OcProgress/OcProgress.vue
+++ b/src/components/atoms/OcProgress/OcProgress.vue
@@ -34,7 +34,11 @@ export default {
      */
     max: {
       type: Number,
-      required: true,
+      required: false,
+      default: undefined,
+      validator: value => {
+        return value >= 0
+      },
     },
     /**
      * The size of the progress bar.
@@ -76,6 +80,9 @@ export default {
       return `oc-progress oc-progress-${this.size} oc-progress-${this.variation}`
     },
     progressValue() {
+      if (!this.max) {
+        return "-"
+      }
       const num = (this.value / this.max) * 100
       return `${num}%`
     },
--- a/src/components/atoms/OcProgress/OcProgress.vue
+++ b/src/components/atoms/OcProgress/OcProgress.vue
@@ -37,7 +37,7 @@ export default {
       required: false,
       default: undefined,
       validator: value => {
-        return value >= 0
+        return value > 0
       },
     },
     /**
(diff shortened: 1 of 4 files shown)


src/components/atoms/OcProgress/OcProgress.vue
```diff
@@ -1,14 +1,17 @@
 <template>
-  <progress
-    :value="value"
-    :max="max"
+  <div
```
pascalwengerter, line 2: Adding a proper review later, but already pretty sure this misses `role="progressbar"` or sth of the likes

JammingBen: 👍

src/components/atoms/OcProgress/OcProgress.vue
```diff
@@ -57,12 +61,24 @@ export default {
...
+    },
   },
 
   computed: {
-    $_ocProgress_classes() {
+    classes() {
       return `oc-progress oc-progress-${this.size} oc-progress-${this.variation}`
     },
+    progressValue() {
+      const num = (this.value / this.max) * 100
```
kulmann, line 79: If `max` can be `null` or `undefined` that edge case should be handled here. E.g. an early return like this:
```js
if(!this.max) {
  return '-'
}
```

kulmann (changes requested):
small nitpicks about the `max` prop. in general awesome, looking forward to seeing it integrated in web 🤩

src/components/atoms/OcProgress/OcProgress.vue
```diff
@@ -30,7 +34,11 @@ export default {
      */
     max: {
       type: Number,
-      required: true,
+      required: false,
+      default: undefined,
+      validator: value => {
+        return value >= 0
```
kulmann, line 40: Doesn't behave correctly:
- `undefined` is valid
- 0 is invalid

kulmann: Huh? `undefined > 0` evaluates to `false` for me

JammingBen: > maybe it doesn't trigger the validator at all since it's the default. that might be...

Yes that's what I meant :D The check itself would return `false` of course

kulmann (approved):
nice 🤩

PR #2165, opened 2022-05-24
Performance improvements in OcTable

pascalwengerter:
## Description
~Introduce `v-once` and cut unnecessary events and Vue.js component usage to complement https://github.com/owncloud/web/pull/7056~

I've removed some OcTable helper components in favor of native HTML table tags - this should reduce rerendering/observing/general UI overhead. It also removes the `shimmer` effect on table cell load (previews get fetched lazily anyways)

PR #2216, opened 2022-06-22
Change remove click event oc icon

pascalwengerter:
## Description
Extracted from https://github.com/owncloud/owncloud-design-system/pull/2165

### DIFF
--- a/changelog/unreleased/change-remove-click-event-oc-icon
+++ b/changelog/unreleased/change-remove-click-event-oc-icon
@@ -1,0 +1,6 @@
+Change: Remove click event on OcIcon
+
+We have removed the default click handler on the OcIcon component, expecting it to increase performance of the UI.
+Adding a custom click-event handling on an instance of the component is of course still possible - albeit then wrapping it in an OcButton is probably the right choice from an accessibility standpoint.
+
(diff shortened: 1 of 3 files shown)


changelog/unreleased/change-remove-click-event-oc-icon
```diff
@@ -0,0 +1,6 @@
+Change: Remove click event on OcIcon
```
kulmann, line 1: IMO not a change. Before this PR the click handler on ocIcon would be used like this:
```js
<oc-icon @click="someCallback" />
```

Since the click handler was living on the root component of the template, writing the code above would still emit a click event like before, no?

pascalwengerter: You can still add an event handler like the `someCallback` example you provided. In theory it's a change because someone somewhere out there might have something listening on an OcIcon emitting a click event, but can re-scope the changelog item if necessary

kulmann: But the only way to listen to that `click` event is by adding a click handler like in the example above. Or could also be that I'm just missing something.

kulmann (approved):
👍

Issue #2238, opened 2022-07-06
Nested ocDrop: both parent and child drop get hidden on show


elizavetaRa:
### Steps to reproduce
1. Create ocDrop inside ocDrop element

### Expected behaviour
Nested ocDrop shows together with parent drop

### Actual behaviour
Both drops get hidden ( tested with all 3 modes)

### Screenshots (if appropriate):
![bug_ocDrop](https://user-images.githubusercontent.com/7430156/177501329-c1966f7f-6f09-4f53-bc72-3c8c14617b7b.gif)

PR #2239, opened 2022-07-06
Nested drop functionality

elizavetaRa:
## Description
Adds "isNested" property to ocDrop that prevents the parent drop from hiding (not fires hideAll() in onShow() of child drop)

## Related Issue

- Fixes https://github.com/owncloud/owncloud-design-system/issues/2238

## Motivation and Context
Useful to create submenus

ocDrop

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [] Documentation added/updated

### DIFF
--- a/changelog/13.1.0_2022-06-07/enhancement-add-nested-drop
+++ b/changelog/13.1.0_2022-06-07/enhancement-add-nested-drop
@@ -1,0 +1,6 @@
+Enhancement: Add nestedd drop functionality
+
+We've added the property "isNested" to ocDrop that prevents the parent drop from hiding by not firing hideAll() in the child drop
+
+https://github.com/owncloud/owncloud-design-system/issues/2238
+https://github.com/owncloud/owncloud-design-system/pull/2239
--- a/src/components/atoms/OcDrop/OcDrop.vue
+++ b/src/components/atoms/OcDrop/OcDrop.vue
@@ -83,6 +83,13 @@ export default {
       required: false,
     },
     /**
+     * Defines if the drop should be nested drop in another drop.
+     */
+    isNested: {
+      type: Boolean,
+      required: false,
+    },
+    /**
      * Element selector used as a target of the drop
      */
     target: {
@@ -143,7 +150,7 @@ export default {
     if (!to || !content) {
       return
     }
-
+    let isNested = this.isNested
     const config = {
       trigger: this.triggerMapping,
       placement: this.position,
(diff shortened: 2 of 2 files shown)


src/components/atoms/OcDrop/OcDrop.vue
```diff
@@ -155,9 +162,11 @@ export default {
       aria: {
         content: "describedby",
       },
-      onShow(instance) {
-        hideAll({ exclude: instance })
-      },
+      ...(!isNested && {
```
kulmann, line 165: Why not like this?
```suggestion
      ...(!this.isNested && {
```

kulmann (commented):
Hm 🤔 Wouldn't that keep other opened OcDrop instances open as well if you open any other OcDrop that has `isNested=true`?

elizavetaRa:
> Hm thinking Wouldn't that keep other opened OcDrop instances open as well if you open any other OcDrop that has `isNested=true`?

@kulmann The parent  (grand...parent by nesting with more levels) ocDrop should not have "isNested" then all the other ocDrop insances get closed on show of the parent. 

I checked as alternative way if hideAll() has an option to exclude more tippy-elements, then it would be possible to hide all but current element and it's parent. But it seems like not. 

If you have any suggestions how to implement submenus otherwise, I am glad to do it :) We at CERNBox need this functionality to show several links in user menu under menu item "Useful links"

src/components/atoms/OcDrop/OcDrop.vue
```diff
@@ -143,7 +150,7 @@ export default {
     if (!to || !content) {
       return
     }
-
+    let isNested = this.isNested
```
kulmann, line 153: line can be removed, now that you use `this.isNested` below

kulmann (changes requested):

kulmann:
> @kulmann The parent (grand...parent by nesting with more levels) ocDrop should not have "isNested" then all the other ocDrop insances get closed on show of the parent.

Well... true 😅 This is fine then. :-) Had one more comment about one line that can be removed now. Otherwise good 👍

kulmann (approved):
👍

changelog/13.1.0_2022-06-07/enhancement-add-nested-drop
```diff
@@ -0,0 +1,6 @@
+Enhancement: Add nestedd drop functionality
```
pascalwengerter, line 1: Sry, could you move this to the `unreleased` changelog folder?

pascalwengerter (commented):
Found a minor changelog thing which I'll fix after merging

PR #2299, opened 2022-08-23
use single-storey letter a

tbsbdr:
use single storey letter "a"

what does it do?  
go to https://rsms.me/inter/lab/?feat-cv11=1 and tick "cv 11 (Single-storey a)" and watch how the small letter "a" changes.

What does it improve?  
The rendering of a single storey letter a aligns with our other rounded letters and digits.

## Before

Screenshot 000316@2x

## After (single storey letter "a")

Screenshot 000315@2x

### DIFF
--- a/changelog/unreleased/enhancement-use-inter
+++ b/changelog/unreleased/enhancement-use-inter
@@ -3,3 +3,4 @@ Enhancement: Use Inter font
 We've switched the default font from Roboto to Inter.
 
 https://github.com/owncloud/owncloud-design-system/pull/2270
+https://github.com/owncloud/owncloud-design-system/pull/2299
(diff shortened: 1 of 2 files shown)


kulmann (changes requested):
Huh, I believe this belongs in the `font-face` definition (see inter.css), but unfortunately that is not supported in browsers other than firefox. -.- Here it works. Not too happy with it, but if it's the only way that works then it's fine.

The linter failed with your PR because the `font-feature-settings: "cv11";` is expected before the `font-family`. Then it should work.

Could you please add the PR link to the changelog item `enhancement-use-inter`?

kulmann (approved):

PR #2325, opened 2022-09-19
Enhancement: Give OcModal the option to use OcContextualHelper

lookacat:
## Description
See https://github.com/owncloud/web/issues/6892#issuecomment-1245438400

## Related Issue
https://github.com/owncloud/web/issues/6892

## Motivation and Context
**Why new props instead of using slot:** So we can use it via code without markup

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests updated
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/enhancement-add-contextual-helper-modal
+++ b/changelog/unreleased/enhancement-add-contextual-helper-modal
@@ -1,3 +1,6 @@
-Enhancement Give OcModal the option to use OcContextualHelper
+Enhancement: Give OcModal the option to use OcContextualHelper
 
 We've added the option for OcModal to use OcContextualHelper
+
+https://github.com/owncloud/owncloud-design-system/pull/2325
+https://github.com/owncloud/web/issues/6892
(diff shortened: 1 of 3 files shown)


JanAckermann:
I am not quite sure if we should do this. 
It would be better to use a slot for the message and implement the contextualHelper there.

lookacat:
The slot method is not suitable because we create the modal via code, we have no option there to dynamically provide a template

JanAckermann:
As far as I know it will be rendered in the code via vue template anyways. 
As far as I know, we do the same with text inputs, so it should be quite possible to add a new variable for the ct helper in web. 

It's just not a good approach at the end to add every component we have to the modal it self.

JanAckermann:
Rechecked, the slot won't be used in the app.vue, so I agree that's the easiest solution here.

JanAckermann (approved):

PR #2367, opened 2022-11-17
DO NOT MERGE: Prepare import into monorepo 

dschmidt:
## Description
This PR contains changes required for importing the ODS into  our monorepo.
I don't actually want them merged, it's here just for reviewing convenience. 

I agreed with @kulmann that we wanted to avoid the explicit style imports in the components, but I couldn't figure out how to achieve that with rollup with reasonable effort. As we plan to switch to Vite very soon and it's a[ documented standard feature](https://vitejs.dev/config/shared-options.html#css-preprocessoroptions) there, I'll hopefully get rid of the imports then.

## Related Issue

- https://github.com/owncloud/web/pull/7940


src/system.js
```diff
@@ -1,54 +0,0 @@
...
-    acc[name] = exp
-    return acc
-  }, {})
-
-export * as composables from "./composables"
-export const utils = { getSizeClass }
-export const components = exportContext(require.context("./components/", true, /\.vue$/))
-export const directives = exportContext(require.context("./directives/", true, /\.js$/))
-export const helpers = exportContext(require.context("./helpers/", true, /\.js$/))
-export const mixins = exportContext(require.context("./mixins/", true, /\.js$/))
```
dschmidt, line 30: This got lost, needs porting

src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    Object.values(directives).forEach(d => Vue.directive(d.name, d))
+  }
+}
+
+export const utils = { getSizeClass }
+export * as components from './components'
+export * as composables from './composables'
+export * as directives from './directives'
+export * as helpers from './helpers'
+export * as mixins from './mixins'
```
dschmidt, line 35: I'm not 100% sure this is exactly the same export structure as before but it's standard and we should do it this way IMHO.

Rollup spits out a warning that we are exporting a default and named modules, it says we should add `output.exports: "named"` and that it would disable the warning. Yes, it does get rid of the warning ... and the default export 😒 
As I would like to switch to Vite in a follow up PR I don't care too much about the warning.

kulmann:
The state of this repo that's available in this branch/PR is the state that we imported into the ownCloud Web repo via https://github.com/owncloud/web/pull/7940

As the master branch of this repo is not needed anymore, I will close this PR now. Thanks for the effort :heart:

